PR #41, opened 2018-02-21
Added quit_on_finish option to GUI allow automated testing

sdfgeoff:
Added this option:
finish

This means you can run (for example):
godot tests/test_all.tscn

This can be done via the command line interface (-gexit flag) but I wanted a way to do this via the GUI. If you think this isn't a good thing to add, then feel free to close the pull request.

### DIFF
--- a/addons/gut/gut.gd
+++ b/addons/gut/gut.gd
@@ -37,6 +37,7 @@ extends "res://addons/gut/gut_gui.gd"
 # Editor Variables
 # ###########################
 export var _run_on_load = false
+export var _quit_on_finish = false
 export(String) var _select_script = null
 export(String) var _tests_like = null
 export var _should_print_to_console = true setget set_should_print_to_console, get_should_print_to_console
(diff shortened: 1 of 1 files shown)


bitwes:
If we quit the GUI, how do you know which tests failed/passed?  Maybe there is a setting in Godot, but when tests run too fast, the console puts out a line that says you are sending out too much output.  

If you can get around that, then it might be just as good to have the scene you have Gut in do the exit.  I think I want to close this but I'm gonna think about it a little more.  If all the output can be visible in the Godot console then this could be useful.

sdfgeoff:
I run something similar to:

```
make test:
    godot path/to/scene.tscn > test.log
    ! grep -q "FAILED" test.log
```
It runs godot, runs all the tests, and saves it into a log file. If any tests fails, the makefile halts.
I have not yet experienced issues with too much output I'll have to double check that.

bitwes:
Ahhhh, that makes sense.  I run into the "too much output" when running just the scene through the editor, but never from the command line so this should be fine.  I also think this is a valid way to do things, since people might want to add things to the scene and run in a batch mode.  

Add a description of the setting to the README and I'll merge it.  I should also document that signal as well.  I'll do that.

bitwes:
I'm working on some bug fixes and additions in a 6_1 branch.  I'm going to close this.  Just reopen it against the 6_1 branch and I'll merge it in.  I can take car of the readme changes...I'm making a bunch already anyhow.

PR #52, opened 2018-04-05
Move everything back to where it was before, but "export-ignore" unwanted files

cmfcmf:
Hi @bitwes,
I've recently installed Gut and it is working great so far!
The only hickup was the requirement to un-check unwanted files before installation. I've seen you recently improved the process by moving as many files as possible into a subdirectory. I think there might be an easier way though: You can use the `.gitattributes` file to exclude certain files and folders from GitHub's `.zip` downloads. Cloning using Git is not affected and works as before. If you'd like to try this out, head over to my fork's `export-ignore` branch and use the `Download ZIP` button: https://github.com/cmfcmf/Gut/tree/export-ignore

I couldn't test this using the AssetLib from within Godot, but I suppose Godot is also just using the `Download ZIP` links.

Let me know what you think!


bitwes:
6.2.0 is released and early indicators are very positive!  The download on the release page now only has the addons/gut folder.  I'm awaiting approval for the new version in the Asset Library.  I'll test it once it is live.

cmfcmf:
Thank you for merging! 🎉 
I've just noticed that I didn't update the `CHANGES.md` file and didn't move the license, will you take care of that?

PR #59, opened 2018-05-02
assert_has_editor_property

Myrkrheim:
Hey there.
First off, gotta say this is a fantastic utility you're cooking up here. :+1:

Anyway, I usually start new node scripts thinking about what signals and/or editor properties they might need. So I wanted a counterpart to assert_has_signal. Probably some overlap with assert_get_set_methods, but I don't always need getters/setters and I like to have a check that the property itself exists on the object.

assert_has_editor_property(obj, property_name, type)
Asserts that passed object has an exported property named 'property_name' of built-in type 'type'
```GDScript
class ExportClass:
	export var some_number = 5
	export(PackedScene) var some_scene
	var some_variable = 1

func test_assert_has_editor_property():
	var obj = ExportClass.new()

	gut.p('-- passing --')
	assert_has_editor_property(obj, "some_number", TYPE_INT)
	assert_has_editor_property(obj, "some_scene", TYPE_OBJECT)
	
	gut.p('-- failing --')
	assert_has_editor_property(obj, 'some_number', TYPE_VECTOR2)
	assert_has_editor_property(obj, 'some_scene', TYPE_AABB)
	assert_has_editor_property(obj, 'some_variable', TYPE_INT)
```
Didn't see a need for it myself, but it would be trivial to make assert_has_variable_property as well.

### DIFF
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -90,6 +90,9 @@ func _init_types_dictionary():
 	types[TYPE_COLOR_ARRAY] = 'TYPE_COLOR_ARRAY'
 	types[TYPE_MAX] = 'TYPE_MAX'
 
+const EDITOR_PROPERTY = PROPERTY_USAGE_SCRIPT_VARIABLE | PROPERTY_USAGE_DEFAULT
+const VARIABLE_PROPERTY = PROPERTY_USAGE_SCRIPT_VARIABLE
+
 # Summary counts for the test.
 var _summary = {
 	asserts = 0,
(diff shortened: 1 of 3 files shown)


addons/gut/test.gd
```diff
@@ -338,6 +341,38 @@ func assert_get_set_methods(obj, property, default, set_to):
...
 
+# ---------------------------------------------------------------------------
+# Property search helper.  Used to retrieve Dictionary of specified property
+# from passed object. Returns null if not found.
+# If provided, property_usage constrains the type of property returned by
+# passing either:
+# EDITOR_PROPERTY for properties defined as: export(int) var some_value
+# VARIABLE_PROPERTY for properties definded as: var another_value
+# ---------------------------------------------------------------------------
+func _find_object_property(obj, property_name, property_usage=null):
```
bitwes: I'm not fond of early `return` and `breaks`.  I'd rather see a `while` loop here.

bitwes:
If you can add some example tests to the `test_readme_examples.gd` then I'll take care of adding it to the Methods wiki page.

Myrkrheim:
Updated.
Anything else, let me know.

Myrkrheim:
Thanks for the merge.

Actually went back and forth on that name myself.
Just went with editor_property because in my head I'm thinking "Does this show up in the editor?", not "Did I write an export statement?" Both questions are asking essentially the same thing so should be fine either way. I do understand the want/need to save keystrokes when typing that assertion though. 
So, go ahead and rename if it suits you better.

bitwes:
I went ahead and renamed it to `assert_exports`.  It'll be in the next release which I hope to get out early next week.  I need to "dog food" the new "spy" feature a little.

PR #61, opened 2018-05-12
Set exit code based on failed tests

cmfcmf:
This allows continuous integration services to detect whether or not the tests
ran through successfully.


bitwes:
I like it!  I'm merging this into the 6_4 branch which should be out early next week at the latest.

Issue #64, opened 2018-06-16
New method for float equality


hbergren:
When attempting to use `assert_eq(float1, float2)`, I get an error message like:

` FAILED:  [91.733333] expected to equal [91.733333]`

Now, I know that floats are never exactly equal, but I would like some way to see if they are more or less equal.

My suggestion is either:
1) A new method called `assert_floats_eq( )` specifically for floats, with a a call signature of `(got, expected, plus_or_minus, message)` where `plus_or_minus` is the acceptable error
2) Add an optional `plus_or_minus` parameter to the standard `assert_eq()`

For now, I can workaround this using an `assert_between` in each of my tests, but I would really like an `assert_approximately_eq` or equivalent.

cmfcmf:
Hi @hbergren,
did you try to push the branch directly to this repository? That won't work, because only @bitwes has permissions to push to this repository directly. You most likely need to fork this repository using the `Fork` button above, then push your branch to your fork, and then create the PR from your fork to this repository. This is also described here: https://guides.github.com/activities/forking/

hbergren:
Hi @cmfcmf ,

Thanks for the quick reply. I'll give that a try a little later. My current WIP PR doesn't yet meet all the requirements listed in the Contribution Guidelines.

PR #65, opened 2018-06-16
Assert Almost Equals

hbergren:
This is my suggested implementation for https://github.com/bitwes/Gut/issues/64, so we have an example to discuss.

I think it would primarily be useful for comparing floats, and things that contain floats (like Vectors).

### DIFF
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -237,6 +237,29 @@ func assert_ne(got, not_expected, text=""):
 			_fail(disp)
 		else:
 			_pass(disp)
+
+# ------------------------------------------------------------------------------
+# Asserts that the expected value almost equals the value got.
+# ------------------------------------------------------------------------------
+func assert_almost_eq(got, expected, error_interval, text=""):
+	var disp = "[" + str(got) + "] expected to equal [" + str(expected) + "] +/- [" + str(error_interval) + "]:  " + text
+	if(_do_datatypes_match__fail_if_not(got, expected, text) and _do_datatypes_match__fail_if_not(got, error_interval, text)):
+		if(got < (expected - error_interval) or got > (expected + error_interval)):
+			_fail(disp)
+		else:
+			_pass(disp)
+
+# ------------------------------------------------------------------------------
+# Asserts that the expected value does not almost equal the value got.
+# ------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


bitwes:
Sorry for the late reply, I've been away and not watching emails and such the past week.  I like this idea.  I'm not sure about the name.  Maybe `assert_almost_equals` or `assert_within_bounds`.  I'll ponder it, review the PR and post feedback in the next day or so.

Thanks for the PR.

bitwes:
Ok, after a little searching I think assert_almost_equals is probably the best name.  Python's unit test framework has it and since gdscript is modeled after Python that will work for me.  If you rename the methods I'll merge it.  

I'm going to mix this in with the next release which will come out in the next week. I have a few ideas for things I might want to get into the next release.  If I get them done I'll merge them in with this change, otherwise I'll make this the only thing in the release.  When you're done I can edit the PR to go agains the "next_release" branch or you can do it.

Sorry again for the delay and thanks for the PR!

addons/gut/test.gd
```diff
@@ -237,6 +237,18 @@ func assert_ne(got, not_expected, text=""):
 			_fail(disp)
 		else:
 			_pass(disp)
+
+# ------------------------------------------------------------------------------
+# Asserts that the expected value roughly equals the value got.
+# ------------------------------------------------------------------------------
+func assert_roughly_eq(got, expected, error_interval, text=""):
```
Schweigert: Please, add `assert_roughly_nq` method.

Schweigert:
Hello! Glad to see this feature! She will be very helpful! However, by default we will not have the same method negation. It would be cool to have this denial!

addons/gut/test.gd
bitwes: I like the idea of an inverse of this assert.  I requested the name change to `assert_almost_eq`.  So the inverse of this would be `assert_almost_ne`.  I'm not sure I like that but I can't think of a better name.

If you can get this into the PR that would be great.  If not then I'll add it to the issue I'm about to add that does the same thing for `assert_between`.

Schweigert: Yes, I liked your name suggested.

hbergren:
Added `assert_almost_ne`, squashed down and rewrote commits. All tests pass

PR #67, opened 2018-06-29
Set exit code on failed tests despite `should_exit` option

Xrayez:
When running gut from command line, it becomes possible to accidentaly
push changes that didn't pass the tests when using continuous integration
if `should_exit` option is not set.

This fixes the described issue by ensuring that the exit code is set regardless
of possible options. It's also not necessary to set exit code to zero because
the program exits with zero by default.

### DIFF
--- a/addons/gut/gut_cmdln.gd
+++ b/addons/gut/gut_cmdln.gd
@@ -396,9 +396,7 @@ func _init():
 
 # exit if option is set.
 func _on_tests_finished():
+	if(_tester.get_fail_count()):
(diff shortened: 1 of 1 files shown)


bitwes:
Good eye.  Merging into "next_release" branch which should be released at the end of the weekend.

cmfcmf:
> It's also not necessary to set exit code to zero because
> the program exits with zero by default.

You're right. I initially decided to set it to `0` explicitly for a bit more verbosity, but I can also see that it might seem superflous to do so.

PR #71, opened 2018-07-01
Add maximize option to command line and the editor

Xrayez:
This fits test runner window to the viewport. The fullscreen mode can be
toggled from the editor, from a command line and be fetched from the gut
configuration file.

Also defined `title_offset` variable to automatically fetch the window's
title bar height that was hardcoded and used for positioning the window.

### DIFF
--- a/.gutconfig.json
+++ b/.gutconfig.json
@@ -1,5 +1,6 @@
 {
   "dirs":["res://test/unit/", "res://test/integration/"],
+  "should_maximize":false,
   "should_exit":true,
   "ignore_pause":true,
   "log": 1,
--- a/addons/gut/gut.gd
+++ b/addons/gut/gut.gd
@@ -35,6 +35,7 @@ extends "res://addons/gut/gut_gui.gd"
 # ###########################
 # Editor Variables
 # ###########################
+export var _should_maximize = false setget set_should_maximize, get_should_maximize
 export var _run_on_load = false
 export(String) var _select_script = null
 export(String) var _tests_like = null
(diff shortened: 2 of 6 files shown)


bitwes:
I like this, but I think it would be better if it was presented as maximizing the window.  Here's the changes I'd like to see.  My justification for the changes is after the list.

* Change the flag from `_fullscreen` to `_should_maximize`.  
* Change `set_fullscreen` to `maximize_window`.  It shouldn't take any parameters, it just does what it does.
* Change the setter to `set_should_maximize` and just have it set the flag.
* Change the getter to `get_should_maximize` and just have it return the value of the `_should_maximize`.  I can't think of a scenario where it would matter if the window was already maximized.  It's ok to maximize a window that is already maximized.  If you can think of a scenario where we would want to check first feel free to add a `is_window_maximized` method that performs the same logic your getter does.  I'm not fond of `return` statements in the middle of methods, so change that method so it only has one `return`.  
* In `_ready` check the flag and then call `maximize_window()` if it is `true`. 

This separates the functionality from the flag.  We are really just setting a property that states if the window should maximize when it runs which is different than actually maximizing the window.  This way, later, if we decide to add a maximize button to the window we an just call `maximize` when clicked.

bitwes:
After thinking about this a little more, I can't see another meaning for "maximize" so we can name the methods `maximize` instead of `maximize_window` and `is_maximized` if that method is needed.

Xrayez:
Oh yeah, when I made this PR I thought `maximize` would be more sensible afterwards.

Reordered the `should_maximize` option to be next to `should_exit` option and apply this option early on.

You're right about that the `is_window_maximized` is kinda useless here, thought it wouldn't hurt but no need to clutter the codebase then.

Issue #73, opened 2018-07-06
Exporting project no longer runs tests


mashumafi:
I made a simple project with examples. Test run fine in editor and the popup is shown for my exported game. The problem is that no tests are detected. Nothing gets run and I am able to preload the test scripts but they still not detected by the popup.

bitwes:
I think there is a flag (or was) to not export scripts as binary or something of that flavor.  This might help.  I was able to get tests to run on an iPhone at one point but I don't know if that still works.  This has been in the back of my head for awhile but haven't got around to figuring out how to get around it.

I believe it is because it has to parse the tests but when it does it doesn't find anything to test.

What platform were you exporting for?

mashumafi:
I am using version 3.0.2 and I tested on windows + Linux/x11 exports.

bitwes:
I'll toy around with this and see if I can't figure out what is going on.  I think it would be nice to be able to run tests on an exported game but not totally necessary.  Your tests shouldn't get deployed but it is nice to be able to put out a test build and run the tests on multiple platforms w/o having to install the editor.

bitwes:
Here's what I've found:

The test scripts aren't run b/c on an export the `.gd` scripts are compiled into `.gdc` files.  A `.gd..remap` file is also created for each `.gd` script.  This is a `ConfigFile` that points to the `.gdc` file.  I assume this is how godot finds the new scripts.

We cannot change Gut to just follow this chain and load the scripts b/c you cannot get a list of Inner Test Classes by inspecting either the script object or the instance of the script.  There is a `script/source` property on a script object but it is null after being exported.  I can't find the names of any Inner Classes in `get_method_list`, `get_property_list`, or `get_meta_list` on either the script or an instance (`script = load('res://somescript.gd')` and instance = `script = load('res://somescript.gd').new()`)

As a workaround I went hunting for an old 2.x setting that allowed you to export scripts as text but it doesn't seem to exist anymore.  One forum post put out the idea to manually remove all the `.gd.remap` and `.gdc` files and put the .gd scripts back but that doesn't seem like a viable solution when working with phones.  It'd be too tedious, especially with android where it's only a click to get your program running on the phone.  https://godotengine.org/qa/30675/can-gd-be-exported-as-text-not-binary-in-3-x

[...]

PR #77, opened 2018-07-10
Allow testing in exports

mashumafi:
This fix is releated to issue #73.
* dir_exists returns false on directories in exports, using return code from open instead
* the export includes .gdc and .gc.remap files, excluding .remap files
* add a 'parser' for .gdc files, it only looks for test functions but not their line numbers

### DIFF
--- a/addons/gut/gut.gd
+++ b/addons/gut/gut.gd
@@ -661,9 +661,8 @@ func add_script(script, select_this_one=false):
 # ------------------------------------------------------------------------------
 func add_directory(path, prefix=_file_prefix, suffix=_file_extension):
 	var d = Directory.new()
-	if(!d.dir_exists(path)):
+	if d.open(path) != OK:
 		return
-	d.open(path)
 	d.list_dir_begin()
 
 	# Traversing a directory is kinda odd.  You have to start the process of listing
@@ -675,7 +674,7 @@ func add_directory(path, prefix=_file_prefix, suffix=_file_extension):
 		full_path = path + "/" + thing
 		#file_exists returns fasle for directories
 		if(d.file_exists(full_path)):
-			if(thing.begins_with(prefix) and thing.find(suffix) != -1):
+			if(thing.begins_with(prefix) and thing.find(suffix) != -1 and thing.find(".remap") == -1):
 				add_script(full_path)
 		thing = d.get_next()
 	d.list_dir_end()
--- a/addons/gut/test_collector.gd
+++ b/addons/gut/test_collector.gd
@@ -52,6 +52,20 @@ var _test_prefix = 'test_'
 var _test_class_prefix = 'Test'
 
 func _parse_script(script):
+	if script.path.ends_with(".gdc"):
+		_parse_gdc_script(script)
+	elif script.path.ends_with(".gd"):
+		_parse_gd_script(script)
+
+func _parse_gdc_script(script):
+	var test = load(script.path).new()
+	for method in test.get_method_list():
+		if(method["name"].begins_with(_test_prefix)):
+			var new_test = Test.new()
+			new_test.name = method["name"]
+			script.tests.append(new_test)
+
+func _parse_gd_script(script):
 	var file = File.new()
 	var line = ""
 	var line_count = 0


addons/gut/gut.gd
```diff
@@ -661,9 +661,8 @@ func add_script(script, select_this_one=false):
 # ------------------------------------------------------------------------------
 func add_directory(path, prefix=_file_prefix, suffix=_file_extension):
 	var d = Directory.new()
-	if(!d.dir_exists(path)):
```
mashumafi: This function was returning false on exports, not sure why

addons/gut/gut.gd
```diff
@@ -675,7 +674,7 @@ func add_directory(path, prefix=_file_prefix, suffix=_file_extension):
 		full_path = path + "/" + thing
 		#file_exists returns fasle for directories
 		if(d.file_exists(full_path)):
-			if(thing.begins_with(prefix) and thing.find(suffix) != -1):
+			if(thing.begins_with(prefix) and thing.find(suffix) != -1 and thing.find(".remap") == -1):
```
mashumafi: exporting creates .remap files for .gd files, we should skip them

addons/gut/test_collector.gd
```diff
@@ -52,6 +52,20 @@ var _test_prefix = 'test_'
 var _test_class_prefix = 'Test'
 
 func _parse_script(script):
+	if script.path.ends_with(".gdc"):
+		_parse_gdc_script(script)
+	elif script.path.ends_with(".gd"):
```
mashumafi: this could be backward-incompatible

addons/gut/test_collector.gd
```diff
@@ -52,6 +52,20 @@ var _test_prefix = 'test_'
...
+		_parse_gd_script(script)
+
+func _parse_gdc_script(script):
+	var test = load(script.path).new()
+	for method in test.get_method_list():
+		if(method["name"].begins_with(_test_prefix)):
+			var new_test = Test.new()
+			new_test.name = method["name"]
```
mashumafi: is there a way to gather more info on the function?

bitwes:
I'll take a look this weekend.  I'd like to toy with your solution a bit.  I'd also like to add some exported tests to the repo for parsing tests.  This is gonna be great when it gets in.  Thanks for looking into this.

bitwes:
I pulled down your branch had tested it out a bit.  I found a bug and in looking into things I've come up with a few thoughts.

## Other Approaches

### Option 1:  Export .gd files as text
There used to be an option to export your .gd files as text instead of binary.  I searched around for it and couldn't find it anywhere.  I posted a question on the forum about it.  If that option still exists then working it into your debug export profile (which I think is a thing you can do) would be the best approach since it won't require any code changes.

https://godotengine.org/qa/30675/can-gd-be-exported-as-text-not-binary-in-3-x

## Option 2:  Keep an index of the test scripts
I think your approach is better than this one but if yours ends up to be impossible for some reason, this would be a fallback.

One approach I that has popped into my head in the past was to create an index of all the tests that were found when Gut ran last and save that to a file which can later be used to load tests in an exported project.

At a high level this is what I think it might look like
* Move the directory traversal logic out of `gut.gd` and into the `test_collector`.
* Add save/load to the `test_collector` so all the tests found can be written to a file.
* Write the file every time Gut runs.
* When exported we load up the file full of tests instead of looking for them in directories.  We'd have to look for .gdc files or .remap files or something to figure it out.  We'd also have to follow the .remap files to the .gdc files (as I describe below).
* Most all of this can probably exist in the `test_collector` and be transparent to `gut.gd`.
* I think the `test_collector` will have to change so that it does a full parsing of the scripts, ignoring the flags that shorten the run list, to write the file but only return things that match the flags.


* Change `gut.gd` to get the tests from the tree-like object instead of looking for them in directories.  Gut 

## Continuing with this approach
If the gdscript files can't be exported as text then we need to address the following things.

### All tests must pass
I ran some of the test scripts on my phone and not everything was passing.  I didn't look into them but all the tests in `test/unit` and ` test/integration` (that should pass) must pass when exported.

### Finding the test scripts
It looks like the .remap files are config files that have a path to where the new binary file is.  I'm not sure why this is done but we should use that system to future proof this.  Instead of ignoring the .remap files and opening the .gdc files we should open the .remap file, find where it points to and open that file as a test script.

I think we can ignore the edge case of both the .remap and the .gd file existing in the same directory.  This shouldn't happen and if it did it would just end up running the same tests twice.

I think the `add_directory` method could use some boy scouting.
`if(thing.begins_with(prefix) and thing.find(suffix) != -1 and thing.find(".remap") == -1):`
should be something like
`if(thing.begins_with(prefix) and (thing.ends_with(suffix) or thing.ends_with(suffix + ".remap"))):`
also, `thing` should probably become `fs_item` for "file system item" or something other than `thing`.  "thing" is _almost_ justifiable here, but it's a horrible name and I'm sorry you had to see that.

### Inner Classes aren't running
The Inner Test classes aren't being found by the test_collector since it actually parses out the word "class" from the .gd file.  It should be possible to get a list of classes in the script but I haven't looked into how to do that yet.  This would be a show stopper for this if it cannot be done.

### Add tests
We need to add tests for all these conditions.  We'll need some samples added to `parsing_and_loading_samples`.  This would include .remap files and the .gdc files.  The test cases should include all the various scenarios where tests can be defined.

bitwes:
I've looked into it a little more and I can't find any way of getting the Inner Class names out of a script.  They aren't included in `get_method_list`, `get_property_list`, or `get_meta_list`.  There doesn't seem to be any "list" that has them.  Since we can't get the list of Inner Classes, we can't continue with this approach.  

I also couldn't find any way to export the gd files as text.  I think the only feasible approach is to create an index of test scripts and Inner Classes.  If it weren't for the Inner Classes we could just use the .remap files.

I'm going to add all this to the Issue and close this PR.

PR #124, opened 2019-06-18
Add assert_freed & assert_not_freed methods

CodeDarigan:
Adds tests to explicitly check to see if an Object has or has not been deleted from memory. Will only work on objects with manual memory management (non-reference types). 

Also adds an extra folder for the related tests. I didn't know where to put this so I just dumped them here to let you decide post merge.


bitwes:
I've seen this but haven't had a chance to review it except in my phone yet.  It's at the top of the list to get this in.  Thanks for the PR.

PR #131, opened 2019-10-08
Fix 130

mschulzeLpz:
Added a small fix to determine the line number for failin test case.

Need help to add a test coverage

### DIFF
--- a/addons/gut/gut.gd
+++ b/addons/gut/gut.gd
@@ -643,9 +643,7 @@ func _pass(text=''):
 func _fail(text=''):
 	_gui.add_failing()
 	if(_current_test != null):
-		var line_text = '  at line '
-		var line_number = _extractLineNumber( _current_test)
-		line_text += str(line_number)
+		var line_text = '  at line ' + str(_extractLineNumber( _current_test))
 		p(line_text, LOG_LEVEL_FAIL_ONLY)
 		# format for summary
 		line_text =  "\n    " + line_text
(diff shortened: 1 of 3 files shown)


addons/gut/gut.gd
```diff
@@ -643,18 +643,28 @@ func _pass(text=''):
...
+func _extractLineNumber(current_test) -> int:
+	var stackTrace:Array = get_stack()
+	if(stackTrace!=null):
+		for index in stackTrace.size():
+			var line:Dictionary = stackTrace[index]
+			var function = line.get("function")
+			if function == current_test.name:
+				return line.get("line")
+	# fallback to original line
+	return current_test.line_number
```
bitwes: I'm not a fan of early `return`s in general.  At first glance it looks like this will always return `current_test.line_number` and that kind of thing can make debugging difficult down the road.  Change this method to do something like:
```
var to_return = current_test.line_number
...
                              to_return = line.get("line")
return to_return
```

bitwes: For consistency, I would like you to remove the explicit `typing` of `stackTrace` and `line`.  It doesn't hurt anything, but it is not used anywhere else in GUT.  If I make the switch to adding a `type` to the variables I'd like to have it consistent across an entire file at least.

mschulzeLpz: done

mschulzeLpz: done

PR #137, opened 2019-12-26
Add assert_connected

### LINKED ISSUE #113: Connection based asserts
Add asserts that verify an object is connected to a signal on another object.  It should be able to verify that an object is connected to another object without having to know the method that it uses for the connection.  The methods used for connections are usually private and it is helpful to verify that an object is "just" connected, and not the details of the connection.

This ordering of parameters is backwards from how `connect` is used but it makes more sense to me grammatically.  The naming and ordering of parameters is up for discussion.
```
assert_connected(obj_that_connects, obj_connected_to, signal_name, (optional)method_name)
```
Example
```
button.connect('pressed', self, '_on_button_pressed')
assert_connected(self, button, 'pressed')
assert_connected(self, button, 'pressed', '_on_button_pressed')
```
If the test fails it should list any connections between the objects to aid in debugging.  In the example above it would list connections between `self` and `button` and vice versa.  It may also be a good idea to list all connections that `self` has if that is possible.

There should be an `assert_not_connected` as well.

ssd71:
Fixes #113 
Please take a look :)

### DIFF
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -514,6 +514,30 @@ func watch_signals(object):
 	_signal_watcher.watch_signals(object)
 
 # ------------------------------------------------------------------------------
+# Asserts that an object is connected to a signal on another object
+# 
+# This will fail with specific messages if the target object is not connected
+# to the specified signal on the source object.
+# ------------------------------------------------------------------------------
+func assert_connected(target_object, source_object, signal_name, text="", method_name=""):
+	var method_disp = ''
+	if (method_name != ""):
+		method_disp = str(' using method: [', method_name, '] ')
+	var disp = str('Expected object ', target_object,\
+		' to be connected to signal: [', signal_name, '] on ',\
+		source_object, method_disp, ':', text)
+	if(method_name != ""):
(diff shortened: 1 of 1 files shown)


bitwes:
I was waiting on some tests.  Since I didn't see any I created a new branch based on the `next_release` branch and pulled in your changes.  I found a couple syntax errors indicating that this wasn't manually tested.  I corrected the errors and did some parameter renaming and minor refactoring (I changed my mind on the ordering of parameters and having a `text` parameter).  I started writing tests that you can use as an example.

There is still some work to be done.  Unfortunately the `get_signal_connection_list` doesn't work as you thought.  I added a comment indicating what is returned.  You have to loop through the thing checking each `source` instead of calling `has`.

If you want to keep working this issue please use the below branch as your base and create your new Pull Requests against that branch.  I'm not going to work on this for a little bit to give you chance to mess with.  Feel free to post any questions.

You can find the branch here:  https://github.com/bitwes/Gut/tree/issue_113_connection_asserts

ssd71:
Understood, thanks! I'll get to working on it in a few hours, sorry about the delay,  had finals week

PR #142, opened 2020-01-24
Fix connection asserts

ssd71:
- Fixed the _is_connected function to check for connections between two objects
- Added more unit tests for assert_connected as well as assert_not_connected

Thank you again for being so patient about this :)
Please review the changes


bitwes:
Well done, thanks for the contribution.  I'll put this in the next point release.

PR #155, opened 2020-04-24
Update gut.gd

astrale-sharp:
Hey,

this greatly increases readability of gut.p(foo) when foo is a really large Dictionary/Array/else

I didnt test it extensivly (which is ironic) but it adds little complexity and should work pretty good

my own test suggests it doesnt break anything ;)

### DIFF
--- a/addons/gut/gut.gd
+++ b/addons/gut/gut.gd
@@ -791,7 +791,18 @@ func _get_files(path, prefix, suffix):
 # printed.
 # ------------------------------------------------------------------------------
 func p(text, level=0, indent=0):
-	var str_text = str(text)
+	var str_text = var2str(text) if typeof(text) in [\
+			TYPE_ARRAY,
+			TYPE_DICTIONARY,
+			TYPE_COLOR_ARRAY,
+			TYPE_INT_ARRAY,
+			TYPE_VECTOR2_ARRAY,
+			TYPE_VECTOR3_ARRAY,
+			TYPE_RAW_ARRAY,
+			TYPE_REAL_ARRAY,
+			] \
+		else str(text)
+			
 	var to_print = ""
 	var printing_test_name = false
 


bitwes:
I missed the notification on this, I'll checkout your PR this weekend and let you know.

bitwes:
Thanks for the contribution.  I like the idea of better formatting but if you didn't want it to do this you would be stuck.  I think having the user call `var2str` themselves offers the most flexibility even if it is annoying.

If you'd like to have this be the default behavior then you could make a wrapper around `test.gd` and have your tests extend that.
```
# res://test/my_custom_test_class.gd
extends 'res://addons/gut/test.gd'

func gut_p(text):
  # do your formatting here
  gut.p(formatted_text)
```

astrale-sharp:
Ah yea, you're right I didn't thought about that!
(its really beautiful though ;), which I could include some pics)

astrale-sharp:
I just remembered why I wanted to do that! If you have very long arrays or dicts, they dont show entirely in Gut.tscn by lack of space, so it might be better to have this behaviour by default and maybe and optional 'raw'  argument for disabling this for user who want to remove it !
what do you think?

bitwes:
Ahhhh, I see.  This is because the control does not wrap.  I've been running at the command line or through VSCode and didn't notice that.  I've also been scrolling with the touchpad and didn't notice that the horizontal scrollbar appears to be busted.  Maybe it should just wrap...the command line does it or add an option to wrap.

bitwes:
Can you try this:  Open up `addons/gut/GutScene.tscn` and find the node `Gut/TextDisplay/RichTextLable` (it is really a TextEdit) and check "Wrap Enabled" and see what you think.

bitwes:
P.S. I added wrapping into the 6.8.3 branch.  I'll dog food it a bit but I think it looks fine.

astrale-sharp:
It ... works!
I still think its prettier with var2str as defaut but I'll keep it too myself :p

PR #158, opened 2020-05-02
Check for object validity earlier in signal watcher

Xrayez:
In Godot 3.2.2-beta I've been getting this error (while `yield_for` test):

```
* test_state
    Testing: test_body_state
    /# Yielding to signal "completed" or for 5 seconds #/ 

Debugger Break, Reason: 'Invalid get index '[Deleted Object]' (on base: 'Dictionary').'
*Frame 0 - res://addons/gut/signal_watcher.gd:152 in function 'clear'
```

I'm not sure what was merged to 3.2.2 that emits this error, yet I've reordered the object validity check and tests pass now. I wonder whether that's a symptom or a solution.

This might be quite related to godotengine/godot#38322, so this might have just revealed a hidden bug in GUT.

Also, consider switching to [`is_instance_valid`](https://docs.godotengine.org/en/3.2/classes/class_@gdscript.html#class-gdscript-method-is-instance-valid) method, works reliably for checking validity in 3.2.

### DIFF
--- a/addons/gut/signal_watcher.gd
+++ b/addons/gut/signal_watcher.gd
@@ -149,8 +149,8 @@ func is_watching(object, signal_name):
 
 func clear():
 	for obj in _watched_signals:
-		for signal_name in _watched_signals[obj]:
-			if(_utils.is_not_freed(obj)):
+		if(_utils.is_not_freed(obj)):
+			for signal_name in _watched_signals[obj]:
 				obj.disconnect(signal_name, self, '_on_watched_signal')
 	_watched_signals.clear()
 


bitwes:
You mentioned that `is_instance_valid` works reliably in 3.2, was `is_instance_valid` flawed in earlier releases?  It looks like it was introduced in 3.1.  I think it would be ok to break backwards compatibility to 3.0 in the next release.  I can make a 3.0 branch and leave it as it is or use `is_instance_valid` in my custom check if the godot version is ok.

Xrayez:
Sorry I didn't mean to say that. `is_instance_valid` has been working good since it was first introduced in 3.1 in my experience. 🙂

Yeah up to you to break compat, I guess most people use at least 3.1 version. It's possible to retain the compatibility later in the existing Utils class though (with `Engine.get_version_info()` or similar), but of course there's a slight performance overhead with having the check the exact version at run-time.

bitwes:
Looks like I've already broken it.  I found some references to `is_instance_valid` already in the code.  Probably thought I found an old thing.  So far nobody has noticed so it's probably fine.  I'll just have to update some documentation.

Can you change PR to be agains the `v6.8.3` branch?

Xrayez:
Changed the base branch to `v6.8.3`.

I guess I have this habit to contribute to `master` branch because that's what Godot Engine repository prefers by default (seen as latest development version there).

PR #164, opened 2020-05-12
Update gitignore

aaronfranke:
The gitignore was missing a few things and was a bit outdated, so I updated it.

* I removed `.fscache` from the gitignore since this file hasn't existed since Godot 2.1.

* I removed `*.import` from the gitignore because these files control import settings.

I also deleted the `.vscode/` folder since that was in the gitignore already, but was committed for some reason, and I committed the two `icon.png.import` files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,20 @@
-.fscache
-.DS_Store
-*.import
+
+# Godot-specific ignores
+.import/
+export.cfg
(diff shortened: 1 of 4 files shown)


bitwes:
Thank you, looks good.  I'll pull this in for release 6.8.3 which should be any day now.

PR #175, opened 2020-06-03
fixed memory leaks

hilfazer:
I was experiencing some orphaned nodes after tests were finished.

There's this comment in the code i don't fully understand:
```
		# This might end up being very resource intensive if the scripts
		# don't clean up after themselves.  Might have to consolidate output
		# into some other structure and kill the script objects with
		# test_script.free() instead of remove child.
		remove_child(test_script)
```
My code frees those test_scripts but i don't know if anything else needs to be done.

### DIFF
--- a/addons/gut/gut.gd
+++ b/addons/gut/gut.gd
@@ -453,6 +453,12 @@ func _end_run():
 	yield(_yield_between.timer, 'timeout')
 	_gui.get_text_box().cursor_set_line(_gui.get_text_box().get_line_count())
 
+	for test_script in _test_script_objects:
+		assert(not test_script.is_inside_tree())
+		test_script.free()
+
+	_test_script_objects = []
+
 	_is_running = false
 	update()
 	_run_hook_script(_post_run_script_instance)
--- a/addons/gut/logger.gd
+++ b/addons/gut/logger.gd
@@ -1,4 +1,4 @@
-extends Node2D
+extends Reference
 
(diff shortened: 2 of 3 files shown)


bitwes:
Your `gut.gd` code works great, unfortunately it didn't work where you had it.  This is due to the way the tests for GUT itself are structured.  There are some tests that end up using `_test_script_objects` to validate what a GUT instance did after all the tests are ran.  I did some more research and the following worked:
```python
func _notification(what):
	if(what == NOTIFICATION_PREDELETE):
		for test_script in _test_script_objects:
			assert(not test_script.is_inside_tree())
			test_script.free()

		_test_script_objects = []

		if(is_instance_valid(_gui)):
			_gui.free()
```
Once I realized this would work, it inspired me to track down more of the orphans and I'm really close to getting rid of all of them.  I just have a few timers left.

Your changes to `logger.gd` and `test_collector.gd` inspired changes in the current branch too.  I ended up just killing the `extends` in `logger.gd`.  `test_collector.gd` got refactored in the `next_release` branch but I had made the same mistake again, so I added your call to `inst.free` into the new code.

Thanks for the PR and the inspiration!  I can't merge this PR but I will make note in CHANGES.md of your contribution.  If you don't have anymore questions, go ahead and close the PR.  

Thanks again.

hilfazer:
Cool. Glad i could help.

PR #198, opened 2020-07-12
exits with 1 if configuration or path is wrong

ceceppa:
After creating a GitHub Action for running GUT tests on GitHub CI I've noticed that the script does exit with error code 0 if the configuration is wrong, or the given path does not exist. And so this causes the CI passing, even tho no test is executed.

This PR address these situations by returning the -1 error code and so making CI fail.

### DIFF
--- a/addons/gut/gut_cmdln.gd
+++ b/addons/gut/gut_cmdln.gd
@@ -346,10 +346,10 @@ func _run_gut():
 			load_options_from_config_file(opt_resolver.get_value('config_file'), opt_resolver.config_opts)
 
 	if(load_result == -1): # -1 indicates json parse error
-		quit()
+		quit(-1)
 	else:
 		if(!all_options_valid):
-			quit()
+			quit(-1)
 		elif(o.get_value('-gh')):
 			print(_utils.get_version_text())
 			o.print_help()
--- a/addons/gut/gut_cmdln.gd
+++ b/addons/gut/gut_cmdln.gd
@@ -346,10 +346,10 @@ func _run_gut():
 			load_options_from_config_file(opt_resolver.get_value('config_file'), opt_resolver.config_opts)
 
 	if(load_result == -1): # -1 indicates json parse error
-		quit(-1)
+		quit(1)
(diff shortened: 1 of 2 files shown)


addons/gut/gut_cmdln.gd
```diff
@@ -346,10 +346,10 @@ func _run_gut():
 			load_options_from_config_file(opt_resolver.get_value('config_file'), opt_resolver.config_opts)
 
 	if(load_result == -1): # -1 indicates json parse error
-		quit()
+		quit(-1)
 	else:
 		if(!all_options_valid):
-			quit()
+			quit(-1)
```
bitwes, line 352: `-1` isn't a valid exit code.  When you pass any value less than 0 to `quit` it will use `OS.exit_code`.  So far GUT only has exit codes for success(0)/fail(1) so change these to be `quit(1)`.

bitwes:
I see you changed the title but the diff still shows `quit(-1)`.  Have you pushed up the changes?  Might as well change "exists" to "exits" in the title too.

ceceppa:
Yes, this morning GitHub was having an issue, so wasn't showing my other push

Xrayez:
On a side note, do you think it would be better to set the exit code non-zero by default, and only set it to zero when tests actually pass? This would be the ultimate solution for CI, but not sure for other cases.

ceceppa:
The exit code I've changed are the only once I've managed to "get wrong" while I was configuring the CI.
Also, I haven't used GUT long enough to know if there are other scenarios where a different exit error is needed, so probably better keep the default to 0?

Xrayez:
Yeah nevermind me, just some food for thought from my side. 🙂

bitwes:
I was tempted by your the idea to default to 1, but I agree it should stay 0 for now.  I don't have any tests for `gut_cmdln.gd` yet.  It's gross to test it.  It might be time to add some to make sure the exit code is always good which would make changing the default a lot easier.  Now that the post-run script has been implemented this might be possible other than writing some in some other language.

Issue #66, opened 2018-06-29
Add `assert_not_between`


bitwes:
I haven't found a need for it but I'm sure there is a use case for asserting a number is not in an arbitrary range of values.  The upper and lower bounds should not be included in the range it is checking against:
* `expected < min and expected > max`

nilold:
Should it be inclusive?
assert_not_between(1,1,5) -> fails
assert_not_between(5,1,5) -> fails

bitwes:
I think exclusive.  I've updated the description to avoid confusion.  My random thought process below...

There appears to be some inconsistencies.  The descriptions said to be inclusive but the code sample did not.  `assert_between` includes the upper bounds but not the lower. I think there was a good reason for doing that but I can't remember it now. `assert_almost_eq` includes low and high.  When I think "pick a number between 1 and 10" then that should be inclusive...and when I think "not between 10 and 20"...I think below 10 and greater than 20...so exclusive. 

I wonder if these should be changed to `assert_within` and `assert_not_within`.  I think "within" implies inclusivity better than "between" does (`assert_within` would be inclusive and `assert_not_within` would not be).  Too late now I guess.  This shouldn't be this hard to figure out.

nilold:
I agree with your semantics. `Within` sounds inclusive and `between`, exclusive, but I don't see this applied in practice. E.g. in SQL, between is inclusive by default. I think the most important is to be clear in the code or documentation.

Another option, which is more user-friendly in my opinion, is to have `assert_not_between_inclusive` and `assert_not_between_exclusive. Doing the same with `assert_between` we would have to have an alias for and maybe deprecate the non-explicit version in the future.
We could do this with between/within as well, but in my opinion, more explicit is better.

This naming would follow the Zen of ~Python~ GdScript. But I would understand if you prefer not to work with deprecation for now.

bitwes:
I'll mull it over a little.  I like to be verbose but prefer shorter names
with tools so I'm on the fence to make two big name asserts   I think the
pass/fail message helps clear things up.  And if you are using this assert
you should be testing the bounds as well.  On the other hand, I've used
this kind of assert for random values so you aren't guaranteed to hit the
bounds of you don't roll enough values.

I've already added a deprecation system to GUT so it's trivial to mark a
method as deprecated.

nilold:
Sure.
[...]

PR #227, opened 2020-10-02
Issue #66 - Implemented assert_not_between

nilold:
This PR is fixing the issue #66 , which is to add an "assert_not_between" test condition.
As discussed in the issue, this condition should probably be exclusive, which is how I implemented in this PR.


bitwes:
Nice work.  Thanks!

PR #243, opened 2020-11-06
Remove more files from ZIP download

clemens-tolboom:
All files apart from `/addons` should be removed from ZIP download.

### Added

`icon.png.import`
`readme_gutconfig.json`

### Changed

All lines got spaces for layout


bitwes (approved):
Good find.

Issue #254, opened 2020-12-18
New version message


bitwes:
Should be able to get the latest version from github and compare against currently running version and print out a message at the end of the run if a new version exists.

bitwes:
latest version number is somewhere in here:
`curl https://api.github.com/repos/bitwes/Gut/releases/latest`

astrale-sharp:
I'd like to do it! :)

It is the last part of the path contained with the key "html_url".
A code like this prints the version number : 
```gdscript
func _ready():
	# Create an HTTP request node and connect its completion signal.
	var http_request = HTTPRequest.new()
	add_child(http_request)
	http_request.connect("request_completed", self, "_http_request_completed")

	# Perform a GET request. The URL below returns JSON as of writing.
	var error = http_request.request("https://api.github.com/repos/bitwes/Gut/releases/latest")
	if error != OK:
		push_error("An error occurred in the HTTP request.")


func _http_request_completed(result, response_code, headers, body):
	var response = parse_json(body.get_string_from_utf8())

	# Will print the user agent string used by the HTTPRequest node (as recognized by httpbin.org).
	var path : String = response.html_url
	var arr = Array(path.split("/"))
	print(arr.pop_back())
```
im working on it, PR should come soon

bitwes:
Ok.  The message should appear at the end of the run.  Any errors should be ignored.

astrale-sharp:
this should be closed now ;)

PR #274, opened 2021-02-10
gut tells you if its up to date

astrale-sharp:
#254

The message appears at the very end of the run
every error is ignored ;)


astrale-sharp:
gut1

gut2

Maybe I should jump a line before the message?

bitwes:
Sweet!  Two small tweaks and I'll start the review

* I think we can  skip telling  the user they are  up  to date.  The less they need to read the better.
* Change the "new version" message to be:
```
GUT version <whatever> is now available.
```

astrale-sharp:
Done :) you want me to squash them in one commit?

bitwes (changes requested):
Break the code out of `_ready` into  a request method called from `_ready`.
There should at least be some tests to make sure that the handling of the response does not  cause errors.  Put these tests in `test/unit/test_utils.gd`.

There are also two bugs that need to be addressed.:

1.  If there is no internet connection then errors are printed in the output:
```
ERROR: _resolve_hostname: getaddrinfo failed! Cannot resolve hostname.
   At: drivers/unix/ip_unix.cpp:112.
[INFO]:  using [/Users/butchwesley/Library/Application Support/Godot/app_userdata/Gut] for temporary output.
Godot version:  3.2.3
GUT version:  7.1.0
ERROR: call: Error parsing JSON at line 0:
   At: modules/gdscript/gdscript_functions.cpp:1268.
```

2.   If the response does  not contain `html_url` then  GUT  hangs.  This should also be ignored.  You can  simulate this with the existing code by taking the last letter off of the url.
```
Debugger Break, Reason: 'Invalid get index 'html_url' (on base: 'Dictionary').'
*Frame 0 - res://addons/gut/utils.gd:121 in function '_http_request_completed'
Enter "help" for assistance.
```

bitwes:
Also add a bullet point to the top of CHANGES.md

astrale-sharp:
I've done some research but im not sure this error can be ignored `ERROR: _resolve_hostname: getaddrinfo failed! Cannot resolve hostname.
   At: drivers/unix/ip_unix.cpp:112.`

there is discussion about this [here](https://github.com/godotengine/godot/issues/22489)

test/unit/test_autofree.gd
```diff
@@ -69,4 +69,4 @@ func test_watch_for_orphans():
 
 func test_watch_for_orphans2():
 	var n = autoqfree(Node.new())
-	assert_false(false)
\ No newline at end of file
+	assert_false(false)
```
bitwes, line 72: revert this change.

bitwes (changes requested):

bitwes (approved):

Issue #259, opened 2021-01-04
Snapshot testing


adrian-goe:
I am currently trying to write a test for a procedural generated map. In order to test this properly, a result is generated that is too large to write by hand or have it written in the testfile.

So something like Snapshot testing would be a nice feature for this.
Example:
https://jestjs.io/docs/en/snapshot-testing

astrale-sharp:
I'm going to try to implement this.
comparing images in godot is fairly simple
```gdscript
func capture_image():
	get_viewport().set_clear_mode(Viewport.CLEAR_MODE_ONLY_NEXT_FRAME)
	# Wait until the frame has finished before getting the texture.
	yield(VisualServer, "frame_post_draw")

	# Retrieve the captured image.
	var img = get_viewport().get_texture().get_data()

	# Flip it on the y-axis (because it's flipped).
	img.flip_y()
	return img

func _path2img(path):
	var img : Image = Image.new()
	img.load("res://my_png.png")
	return img

func _cmp_img(img1, img2) -> bool:
	return img1.get_data() == img2.get_data()
```

the question remaining would be : how do we want to use this from Gut? something like
```gdscript

func test_the_screen_remains_the_same():
	setup_screen()
	#  implicitly captures the screen and compares it to "my_screen_capture.png"
	assert_snapshot_equals("res://my_screen_capture.png")
```

something like that would pretty much work no?
maybe to be able to capture the screen we could expose `capture_image()` so we can access it like this:
```gdscript
var img : Image = _strutils.capture_image()
img.save_png("some_pic.png")
```

what do you think? (im waiting for an answer before i actually start working on gut for this)

adrian-goe:
Images would of course also be a possibility. But I had rather thought of data, e.g. JSON or text. 

Then it would look like this from Gut:
```gdscript
func test_something():
    var worldSize = 100
    var result = WorldGenerator.generate_world(worldSize)

    assert_snapshot_json(result) # or assert_snapshot(result) for raw text?
```
In jest, a `snapshots` folder is created. In this folder, the test data is represented as a folder and when a snapshot is taken, the name of the test function is taken as the file name. If no snapshot is used in the test class, it does not have to be created.
`res://snapshots/worldgeneartor-test/test_something.json`

In the event that I have two snapshots in one function, I would expect that files would be created in this way:
`res://snapshots/worldgeneartor-test/test_something.json`
`res://snapshots/worldgeneartor-test/test_something_1.json`
[...]

PR #277, opened 2021-02-15
starting to implement snapshot

astrale-sharp:
#259 
this should get working soon


astrale-sharp:
Okay so the JSON spec doesn't support storing int : meaning a limitation of the snapshot is that if you really care about type safety between int and floats and want to check it stays the same this wont be supported (I don't think it matters very much though)

bitwes:
Tiding up PRs and issues.  Make another PR if you end up working on this more.

PR #295, opened 2021-05-08
Makes yield between tests using the autofree queue configurable

abaire:
My project has a large number of small tests that use `add_child_autoqfree` and the 0.1 second yield between tests ends up taking more time than the tests themselves (in a subset 5.2 seconds w/ `0.1`, 0.9 with `0.01`). 

On my test machines reducing the yield to `0.01` does not seem to trigger any incorrect orphans but I don't know if this is deterministic in other environments so I made it a configurable parameter.


bitwes:
I've added a `yield_frames` method that counts frames instead of elapsed time.  I've set it to wait 2 frames whenever it would do a "yield between".  I think this should fix things.  It's in master, can you give it a shot?

abaire:
Sorry, missed your comment. My test suite has changed a bit, but the frame-based wait seems to have improved things - thanks!

PR #306, opened 2021-08-30
Add a message to show what scripts are being added in CMDLN

smks:
Nice to see what scripts are going to be tested in isolation
Screenshot 2021-08-30 at 16 07 19

### DIFF
--- a/addons/gut/gut_cmdln.gd
+++ b/addons/gut/gut_cmdln.gd
@@ -283,6 +283,7 @@ func apply_options(opts):
 		_tester.add_directory(opts.dirs[i], opts.prefix, opts.suffix)
 
 	for i in range(opts.tests.size()):
+		print('Adding script: ' + str(opts.tests[i]))
 		_tester.add_script(opts.tests[i])
 
 	if(opts.selected != ''):


bitwes:
I'm not sure about this, I'll have to think about it some more.  You can see the scripts that it ran in the output and any scripts that aren't found are listed at the start of the run.  The "Adding" text doesn't take into account that the file may not exist which could be confusing.

I noticed that all your tests end with `.test.gd`.  Are you adding these tests via the command line because they don't match the existing convention?  

You can change the prefix and suffix for test scripts via the command line (or `.gutconfig.json`).  You'd have to change the prefix to `''` and the suffix to `'test.gd`.  I'm not sure if GUT would like that though.  GUT might be using the built-in `get_extension` on `string` and I don't know how it acts with multiple periods.

smks:
This might be a good feature for GUT, I wrote this bash script in the CI build that runs a stage highlighting if any of your scripts are missing a test file. Could be tweaked in the Godot language.

```
#!/bin/bash

RED='\033[0;31m'
GREEN='\033[0;32m'
NC='\033[0m' # No Color
FILES="${PWD}/src/*"
testFileCount=0
noTestFileCount=0

for file in $FILES
do

  if [[ -d $file ]]; then
    continue
  fi

  if [[ "$file" == *".test"* ]]; then
    continue
  fi

  if [[ "$file" == *".tscn"* ]]; then
    continue
  fi

  testFile="${file//.gd/.test.gd}"

  if [[ ! -f "$testFile" ]]
  then
    echo "${RED}- Missing test file $testFile"
	noTestFileCount=$(( noTestFileCount + 1))
  else
    echo "${GREEN}+ Found \"$testFile\""
	testFileCount=$(( testFileCount + 1))
  fi
done

total=$(( $testFileCount + $noTestFileCount ))

echo "${NC}=========="
echo "${GREEN} + Test file count ${testFileCount}/${total}"
echo "${RED} - NO Test file count ${noTestFileCount}/${total}"
```

bitwes:
I toyed around with this.  It doesn't seem to cause any extra output if you aren't adding them one by one.  So I'm ok with it.  The latest changes to GUT moved this function to `gut_config.gd` line 131.  Create another PR and I'll merge it.

Issue #284, opened 2021-03-16
Run tests in the same order everytime


bitwes:
The order of the following can vary from run to run (due to Godot):
* Files in a directory
* Inner classes in a file
* method names in a class/file

This means that the tests  are executed in a different order just about every run.  This can make comparing output from one run to the next difficult.  

It would be best if the test were run in this order:
 * Alphabetical by file name
 * line number for Inner classes
 * line number for methods

If this isn't feasible then alphabetical for file, inner class, and method should be possible.

If you want to ensure that you aren't building in some test order dependencies then the user should be able to specify a seed to randomly order the files/inner classes/methods.

hilfazer:
> If you want to ensure that you aren't building in some test order dependencies then the user should be able to specify a seed to randomly order the files/inner classes/methods.

I'd also like an option to not specify a seed so the RNG orders tests differently every time i run them.

bitwes:
The next release of GUT has an xml export and some hidden away json export logic.  This should make comparing results easier.  I'm leaving this open because I still want it but I'm taking it out of the 7.2 milestone.

gabahulk:
I find this issue a bit puzzling... Aren't unit tests supposed to be independent? If this issue is implemented, it should be optional and NOT the default.

bitwes:
They are, but having some agency over test execution order is good.  I don't know if test order is random still.  With this feature (including a random option with and without a user supplied seed) you can at least get predictable results as the engine implementation changes.

PR #325, opened 2021-10-26
Ordered Execution

eumario:
This PR will order the methods on a Test Script, based upon where they are declared in the test file.  Fixes bug #284

### DIFF
--- a/.gut_editor_config.json
+++ b/.gut_editor_config.json
@@ -12,7 +12,7 @@
  "hide_orphans": false,
  "ignore_pause": true,
  "include_subdirs": false,
- "inner_class": "",
+ "inner_class": null,
  "junit_xml_file": "",
  "junit_xml_timestamp": false,
  "log_level": 1,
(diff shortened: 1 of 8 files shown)


eumario:
Sorry Butch,

Ended up getting busy with RL, so haven't made much progress on this.  But I have added the Script Parser to the Gut Library, along with unit tests testing the functionality of it, so that part is done.  If you want to take a look at it, feel free to give me some feed back on the system I implemented in the library and tests.

bitwes:
Re: RL, not a problem, I do that on a regular basis.

I checked out the parser and related tests.  It looks like a good start.  I'm assuming you haven't gotten to the inner classes yet.  

A couple style notes.
* All private variables (which in general they all should be) must start with `_`.  So `code_tree` should be `_code_tree`, etc.  This means you can skip adding `= {} setget nosetter` if you want.  With the `_` convention, you know something is being used incorrectly if you see something like `foo._bar = 'foobar'`.  (I may have broken this rule, but I'm the only one allowed to)
* GUT doesn't use any strict-typing.  I can make a concession for new files if it helps significantly.  It's better to keep the code base consistent If you aren't getting any specific benefits from it.    Do not use it in existing files (unless you want to refactor the whole thing, heh).

bitwes:
Tiding up PRs and issues.  Make another PR if you end up working on this more.

PR #328, opened 2021-11-22
Added assert_in and assert_not_in

Erawpalassalg:
Enables assert using the "in" built-in method. Makes assert on
Pool*Array possible.

### DIFF
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -470,6 +470,26 @@ func assert_not_between(got, expect_low, expect_high, text=""):
 				_pass(disp)
 
 # ------------------------------------------------------------------------------
+# Uses the built-in "in" method to determine if the object is contained in
+# the passed in collection.
+# ------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


bitwes:
Is there any difference in how this works and how `assert_has` works?

Erawpalassalg:
This works with `String` and `Pool*Array` objects, which do implement a `has` method.

On the other hand `has` is useful for custom objects (I'm thinking about a `Set` object which seems to be popular) and would probably implement an `has` method but wouldn't work with `in`. For more classic collections such as `Array` and `Dictionary` they would behave the same.

bitwes:
It's been a bit.  If you are still interested in implementing this feature then add in some tests and I'll merge it.  Check some expected passing some conditions, failures, and some invalid inputs.

Erawpalassalg:
Hi, that shouldn't be an issue.

bitwes:
Been a bit.  Closing, reopen if you still want to work on this.

PR #335, opened 2022-01-01
Bug fixed in test.gd: should pass a string to has_method.

strank:


addons/gut/test.gd
```diff
@@ -1112,7 +1112,7 @@ func _assert_setget_called(type, name_property, setter = "", getter  = ""):
 	if name_getter != '':
 		expected_calls_getter = 1
 		stub(obj, name_getter).to_do_nothing()
-		var new_property = obj.get(name_property)
+		var _new_property = obj.get(name_property)
```
bitwes, line 1115: To avoid any confusion between local variables/parameters that are ignored and class level variables, ignored variables and parameters should get two leading underscores.

bitwes (commented):
Rename `_new_property` to `__new_property` and I'll merge this in.  Good find, I have no idea how this ever made it out the door.

addons/gut/test.gd
strank: Sure, done.

strank:
I found this while going through the educational exercise of making Gut run on 4.0 / Godot master.
I'll put my changes into the 4.0 branch of this repo clone when I'm in case you are interested.
Apart from this here, I found two things that might be worth doing even now:

* mixed indentation with tabs and spaces are considered an error in Godot 4 - used in a few places to align continuation lines.
  I'd suggest, just using a standard two-tab indent for continuations.
* Setting variables of built-in types to null is impossible when adding type hints, and sometimes marked as an error otherwise.
  So maybe it is worth to think of other ways of indicating "no value yet", or lobby for the nullable? types proposal that's on godot-proposals.

bitwes:
@strank Merged.  I created two issues and added a new issue tag of Godot 4.0.  Can you add some examples to #339 where this is happening in GUT.  I kinda followed what you are saying but not completely.

PR #348, opened 2022-03-28
Update documentation for assert_is/assert_extends

### LINKED ISSUE #347: `assert_is` says it's alias, but it is not
The documentation for `assert_is` says it is an alias for `assert_extends`.
However, it is the opposite - `assert_extends` is now an alias to `assert_is`.

Expected:
* Documentation for `assert_is` should read:
> Assert that object is an instance of a_class
* Documentation for `assert_is` should read:
> Deprecated. Use `assert_is`.

Structed:
Fixes #347


bitwes:
Works for me.  Thanks.

PR #395, opened 2022-08-16
4.0 Name Change Errors

Tim-Fronsee:
### Summary

Fixes a few errors relating to 4.0 name changes that appear after enabling the plugin within a 4.0 project.

#### Additional Comments

I noticed there are two 4.0 related branches.  Not sure if this is the main one, but seeing as it was updated recently I decided to use it as a base.


bitwes (approved):
Thanks.

PR #398, opened 2022-08-18
4.0 Syntax Highlighting & Label Settings

Tim-Fronsee:
#### Fixes

-  `OutputText` now uses the new `sytnax_highlighter` property for keyword highlighting
- Font color is now accessed from the theme in `OutputText`
- `RunResults` uses `get_label_settings` if available to grab the font from `FontSampler`
- More `rect_size` -> `size` conversions


bitwes (approved):
Looks good.

PR #400, opened 2022-08-20
4.0 DirectoryCtrl::set_text Infinite Loop Fix

Tim-Fronsee:
#### Bug

An infinite loop occurs while attempting to initialize `gut_config_gui` due to `set_text` calling into itself via `text = t`, resulting in an editor crash while opening a project with the plugin enabled.  In case it's relevant, I'm running the editor and plugin on Windows 10.

Interestingly, I received no error messages (such as an out of memory error / exception) from the engine while trying to nail down the source, even when running the editor in verbose mode.  

#### Fix 

I removed the `set_text` function in `DirectoryCtrl`, migrating the logic into the setter. 

~~#### Extras~~

~~I changed the `add_child` calls to `call_deferred` in `DirectoryCtrl::_ready` as I believe this is the preferred method for adding children in `_ready`.~~

### DIFF
--- a/addons/gut/gui/gut_config_gui.gd
+++ b/addons/gut/gui/gut_config_gui.gd
@@ -7,7 +7,9 @@ class DirectoryCtrl:
 		get:
 			return get_text()
 		set(val):
-			set_text(val)
+			text = val
+			_txt_path.text = text
+
 	var _txt_path = LineEdit.new()
 	var _btn_dir = Button.new()
 	var _dialog = FileDialog.new()
@@ -25,7 +27,7 @@ class DirectoryCtrl:
 		_dialog.size = Vector2(1000, 700)
 
 	func _on_selected(path):
-		set_text(path)
+		text = path
 
 
 	func _on_dir_button_pressed():
@@ -34,16 +36,12 @@ class DirectoryCtrl:
 
 
 	func _ready():
-		add_child(_txt_path)
-		add_child(_btn_dir)
-		add_child(_dialog)
+		call_deferred('add_child', _txt_path)
+		call_deferred('add_child', _btn_dir)
+		call_deferred('add_child', _dialog)
 
 	func get_text():
 		return _txt_path.text
-
-	func set_text(t):
-		text = t
-		_txt_path.text = text
 
 	func get_line_edit():
 		return _txt_path
--- a/addons/gut/gui/gut_config_gui.gd
+++ b/addons/gut/gui/gut_config_gui.gd
@@ -5,9 +5,8 @@ class DirectoryCtrl:
 
 	var text = '':
 		get:
-			return get_text()
+			return _txt_path.text
 		set(val):
-			text = val
 			_txt_path.text = text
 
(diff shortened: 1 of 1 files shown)


addons/gut/gui/gut_config_gui.gd
```diff
@@ -7,7 +7,9 @@ class DirectoryCtrl:
 		get:
 			return get_text()
```
bitwes, line 8: Change this to `return _txt_path.text`

addons/gut/gui/gut_config_gui.gd
```diff
@@ -7,7 +7,9 @@ class DirectoryCtrl:
...
 		set(val):
-			set_text(val)
+			text = val
```
bitwes, line 10: Remove the assignment this assignment.  It will make it clear that this is just a wrapper around `_txt_path.text`.

addons/gut/gui/gut_config_gui.gd
```diff
@@ -34,17 +36,13 @@ class DirectoryCtrl:
...
 
 	func _ready():
-		add_child(_txt_path)
-		add_child(_btn_dir)
-		add_child(_dialog)
+		call_deferred('add_child', _txt_path)
+		call_deferred('add_child', _btn_dir)
+		call_deferred('add_child', _dialog)
 
 	func get_text():
```
bitwes, line 43: Delete this method since the `get():` above will return this value.

bitwes (changes requested):
Good find.  I've fixed a number of these in other places.  This new `set`/`get` is nice, but when there is a backing variable it loses its mind.

Although this is an "experiments" branch, it's starting to have enough changes that it might become the real thing.  So, in this case I would rather just make the "text" property be a wrapper around `_txt_path.text` via the `get` and `set` (as per my requested changes).

Thanks for all the PRs.  It has been very helpful.

bitwes:
I forgot to add, do you have any info on the `call_deferred` for adding children in `_ready`?  I have not heard this before.

Tim-Fronsee:
> I forgot to add, do you have any info on the `call_deferred` for adding children in `_ready`? I have not heard this before.

I've seen this error in my own plugin when calling `add_child` from `_enter_tree`, but I was not seeing it from Gut.  

There's an existing [forum post](https://godotengine.org/qa/28050/parent-node-is-busy-setting-up-children-add_node-failed) featuring the error, for which I based my assumption (perhaps incorrectly) regarding `_ready`.  

Admittedly, this change expands the PR beyond addressing the infinite loop.  If you'd like, I will revert the `add_child` changes along with the requested changes.

bitwes (approved):

bitwes:
>There's an existing [forum post](https://godotengine.org/qa/28050/parent-node-is-busy-setting-up-children-add_node-failed) featuring the error, for which I based my assumption (perhaps incorrectly) regarding _ready.

I have seen that error before.  If it shows up then we should address it.  I agree with the revert.

Issue #384, opened 2022-08-03
Godot 4.0 Things


bitwes:
Here's a list of stuff that will have to be addressed, I tired of creating a bunch of issues.

- [x] `OS.get_unix_time()` Does not exist
- [x] `GDScriptFunctionState` changed
- [x] `list_dir_begin` changed
- [x] `JSON.print` doesn't exist, changed to `stringify`
- [x] `parse_json` doesn't exist
- [x] `if(!newone is Reference):` `Reference` does not exist.
- [x] the return of `get_data` in `JSON` has changed to be an Variant or `null`.
- [x] `.pressed` became `.button_pressed` on all buttons.
- [x] `String.find_last` does not exist anymore.
- [x] Cannot use `await` with signals added via `add_user_signal`.  You must use the `signal` keyword.
- [x] Checking a thing to see if it is null by using `!` does not work anymore.  Probably never should have been used, but I evidently did in some places to see if a dictionary was null `if(!_summary):`.
- [x] 'Invalid set index 'exit_code' (on base: 'OS') with value of type 'int'.'
- [x] The auto converter changed `remove` on a `Directory` object to `remove_at` because it thought it was an array or something.
- [x]  `str` on freed objects changed:  [Failed]:  ["[Freed Object]"] expected to equal ["[Deleted Object]"]:  
- [x] The flags for methods have changed.  The double logic has to change to use the new flags or no methods are found.
- [x] Methods local to a script are now retrieved with `get_script_method_list`, they are not in `get_method_list`.  They each appear to have a flag of `1`.  There also appears to be more information about parameters and maybe defaults than 3.x has.
- [x] Calling super's method now requires `super.` instead of just the `.`.  
- [x] Currently cannot double objects that have methods that `await`.  Since the result of the call to super is returned, Godot errors saying that the method is a coroutine and must be called with `await`.  Could just await every call to super.
- [x] `comparator.gd` and `test.gd` have to be updated to know that they can compare `TYPE_STRING` to `TYPE_STRING_NAME`.
- [x] `GDScript.reload` generates `ERROR: Attempt to open script '' resulted in error 'File not found'.` for doubles.  Created issue https://github.com/godotengine/godot/issues/65263
- [x] `inst2dict` renamed to `inst_to_dict`

russmatney:
[...]

PR #409, opened 2022-08-29
Allow simulate to check whether a Node has disabled its process methods

coffeebeats:
Right now, Gut's `simulate` function will call `_physics_process` and `_process` on a `Node` [as long as the methods are overridden](https://github.com/bitwes/Gut/blob/fe292bcdcf9baca6022944ecdb0f973083ff3ed4/addons/gut/gut.gd#L1349-L1352). However, it's possible for a `Node` to disable its `_physics_process` method using `set_physics_process(false)` and `_process` method using `set_process(false)`. The `simulate` function should respect when a `Node` toggles the status of its  `_physics_process` and `_process` methods.

Regarding the implementation here, I think it would be correct to simply replace the `if(obj.has_method("_process")):` check with `if (obj.is_in_group("idle_process")):`. This is because these groups are [set by Godot](https://docs.godotengine.org/en/stable/classes/class_node.html?highlight=set_process#class-node-method-set-process) when the method is defined and updated when the `set_[physics_]process` methods are called. However, in an effort to provide backwards-compatibility, I've used a parameter on `simulate` to allow opting in to this behavior. This should preserve existing behavior in the event that a current user manually modifies the membership of simulated `Node`s in these Godot-added groups.

NOTE: With Godot 4.0+, the name of these groups is changing. I've added a comment to point this out, but perhaps an additional bug can be created (or this can be added to #384 perhaps?).

### DIFF
--- a/test/unit/test_gut.gd
+++ b/test/unit/test_gut.gd
@@ -12,18 +12,6 @@ var Test = load('res://addons/gut/test.gd')
 #--------------------------------------
 class WithoutProcess:
 	extends Node
-	var call_count = 0
-	var delta_sum = 0.0
-
-	func _get(property):
-		match property:
-			"_process":
-				call_count += 1
-				return null
-			"call_count":
-				return call_count
-			_:
-				return null
 
 class WithProcess:
 	extends Node
@@ -40,17 +28,6 @@ class WithProcess:
 #--------------------------------------
 class WithoutPhysicsProcess:
 	extends Node
-	var call_count = 0
-
-	func _get(property):
-		match property:
-			"_physics_process":
-				call_count += 1
-				return null
-			"call_count":
-				return call_count
-			_:
-				return null
 
 class WithPhysicsProcess:
 	extends Node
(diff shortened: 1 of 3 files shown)


bitwes:
Thanks for the PR.  You'll need to add some tests to `test/unit/test_gut.gd` for this feature and address the comments below.  I think this is a fine addition

https://github.com/bitwes/Gut/blob/master/test/unit/test_gut.gd#L224

> Regarding the implementation here, I think it would be correct to simply replace the if(obj.has_method("_process")): check with if (obj.is_in_group("idle_process")):.

Using `simualte` does not require that a node be in the tree.  In fact it is probably better that it is not, so that you don't get any unexpected calls to `_process` or `_physics_process`.  Since groups require that a node be in the tree, this feature should check `is_processing ` and `is_physics_processing` instead of checking the group.  

> However, in an effort to provide backwards-compatibility, I've used a parameter on simulate to allow opting in to this behavior. 

I think that this new way is the expected way.  I think the optional parameter should be enabled by default.  Without this mechanism, you cannot test more complex setups of multiple objects where some may not have `_process` enabled  (though it could be argued that that would be better suited for an integration test with nodes in the tree and using `yields` and the like).

bitwes:
I think the difference in behavior, when in/out of the tree, is too drastic.  If we change the optional flag to be `check_is_processing` (which would apply to `is_physics_processing` as well) we can simplify the behavior and make the usage clearer.  We would default it to `false` so that the we can preserve the existing behavior.  When it is set to `true` then it will check the flags before running `_process` and `_physics_process`.  When `false`, it will always run those methods.  This way we can have the same behavior regardless of whether or not the object being tested is in the tree.

This will make the setup for a test a  little more complicated in some scenarios, but most of the time it won't matter.  The documentation can point out the setup differences when `check_processing` is set to `true`.
* When outside the tree `is_processing` and `is_physics_processing` is false until you change it via `set_process` or `set_physics_process`.
* After being added to the tree the "processing" flags are `true` if the corresponding methods have been overloaded.  You must call `set_process` and `set_physics_process` to disable them. 

Thoughts?

> One question I have about this implementation is whether it still makes sense to check if the object has actually overridden _process

`_process` and `_physics_process` are virtual methods and do not exist unless overridden, so you have to check for both.
```
    var n = Node.new()
    print(n.has_method('_process'))
```
prints "false"

test/unit/test_gut.gd
```diff
@@ -221,44 +250,194 @@ func test_delete_all_files_in_a_directory():
...
+	var without_method = WithoutProcess.new()
+
+	gr.test_gut.simulate(without_method, 5, 0.2)
+	gr.test.assert_eq(without_method.call_count, 0, "_process should not have been called")
+
+	assert_pass(3)
+	with_method.queue_free()
+	without_method.queue_free()
+
+func test_simulate_calls_process_on_descendents_if_objects_have_method():
```
bitwes, line 273: Rename this to something like `test_simulate_checks_process_on_all_nodes`.

test/unit/test_gut.gd
```diff
@@ -221,44 +250,194 @@ func test_delete_all_files_in_a_directory():
...
+	
+	var without_method = WithoutProcess.new()
+
+	gr.test_gut.simulate(without_method, 5, 0.2)
+	gr.test.assert_eq(without_method.call_count, 0, "_process should not have been called")
+
+	assert_pass(3)
+	with_method.queue_free()
+	without_method.queue_free()
+
```
bitwes, line 272: Put the original `test_simulate_calls_process_on_child_objects_of_child_objects` back.

```
func test_simulate_calls_process_on_child_objects_of_child_objects():
	var objs = []
	for i in range(5):
		objs.append(HasProcessMethod.new())
		if(i > 0):
			objs[i - 1].add_child(objs[i])
	gr.test_gut.simulate(objs[0], 10, .1)

	for i in range(objs.size()):
		gr.test.assert_eq(objs[i].process_called_count, 10, "_process should have been called on object # " + str(i))

	assert_pass(objs.size())
```

test/unit/test_gut.gd
```diff
@@ -221,44 +250,194 @@ func test_delete_all_files_in_a_directory():
...
+func test_simulate_calls_process_on_descendents_if_objects_have_method():
 	var objs = []
-	for i in range(5):
-		objs.append(HasProcessMethod.new())
+	for i in 4:
+		# Odd indices will be 'WithoutProcess'
+		objs.append(WithProcess.new() if not i % 2 else WithoutProcess.new())
```
bitwes, line 277: One line is nice, but it took awhile to decipher exactly what it was doing.  The `if not` and the fact that `i % 2` evaluates to `true` when it is `not` divisible by `2` is confusing.  Swap this out for a basic `if`
```
		if(i % 2 == 0):
			objs.append(WithProcess.new())
		else:
			objs.append(WithoutProcess.new())
```

test/unit/test_gut.gd
```diff
@@ -221,44 +250,194 @@ func test_delete_all_files_in_a_directory():
...
+	gr.test.assert_eq(objs[2].call_count, 5, "_process should have been called 5 times")
+	gr.test.assert_eq(objs[2].delta_sum, 1.0, "The delta value should have been passed in and summed")	
+
+	gr.test.assert_eq(objs[3].call_count, 0, "_process should not have been called")	
+
+	assert_pass(6)
+	for obj in objs:
+		obj.queue_free()
+
+func test_simulate_calls_process_if_object_is_processing():
```
bitwes, line 297: This should be 3 tests.  Basically one per variable.

test/unit/test_gut.gd
```diff
@@ -221,44 +250,194 @@ func test_delete_all_files_in_a_directory():
...
+	assert_pass(4)
+	is_processing.queue_free()
+	is_not_processing.queue_free()
+	is_processing_but_without_method.queue_free()
+
+func test_simulate_calls_process_on_descendents_if_objects_are_processing():
+	var objs = []
+	for i in 4:
+		# Odd indices will be 'WithoutProcess'
+		objs.append(WithProcess.new() if not i % 2 else WithoutProcess.new())
```
bitwes, line 326: Do the same thing with this one liner, replace with a basic `if`.

test/unit/test_gut.gd
```diff
@@ -221,44 +250,194 @@ func test_delete_all_files_in_a_directory():
...
+	gr.test.assert_eq(objs[2].call_count, 5, "_process should have been called 5 times")
+	gr.test.assert_eq(objs[2].delta_sum, 1.0, "The delta value should have been passed in and summed")	
+	
+	gr.test.assert_eq(objs[3].call_count, 0, "_process should not have been called")
+
+	assert_pass(5)
+	for obj in objs:
+		obj.queue_free()
+
+func test_simulate_calls_physics_process_if_object_has_method():
```
bitwes, line 349: This should be two tests.

test/unit/test_gut.gd
```diff
@@ -221,44 +250,194 @@ func test_delete_all_files_in_a_directory():
...
+
+	assert_pass(3)
+	with_method.queue_free()
+	without_method.queue_free()
+
+func test_simulate_calls_physics_process_on_descendents_if_objects_have_method():
+	var objs = []
+	for i in 4:
+		# Odd indices will be 'WithoutPhysicsProcess'
+		objs.append(WithPhysicsProcess.new() if not i % 2 else WithoutPhysicsProcess.new())
```
bitwes, line 369: Simple `if` here.

test/unit/test_gut.gd
```diff
@@ -221,44 +250,194 @@ func test_delete_all_files_in_a_directory():
...
+	gr.test.assert_eq(objs[2].call_count, 5, "_physics_process should have been called 5 times")
+	gr.test.assert_eq(objs[2].delta_sum, 1.0, "The delta value should have been passed in and summed")	
+
+	gr.test.assert_eq(objs[3].call_count, 0, "_physics_process should not have been called")	
+
+	assert_pass(6)
+	for obj in objs:
+		obj.queue_free()
+
+func test_simulate_calls_physics_process_if_object_is_processing():
```
bitwes, line 389: This should be 3 tests.

test/unit/test_gut.gd
```diff
@@ -221,44 +250,194 @@ func test_delete_all_files_in_a_directory():
...
+	assert_pass(4)
+	is_processing.queue_free()
+	is_not_processing.queue_free()
+	is_processing_but_without_method.queue_free()
+
+func test_simulate_calls_physics_process_on_descendents_if_objects_are_processing():
+	var objs = []
+	for i in 4:
+		# Odd indices will be 'WithoutPhysicsProcess'
+		objs.append(WithPhysicsProcess.new() if not i % 2 else WithoutPhysicsProcess.new())
```
bitwes, line 418: Simple `if` here.

bitwes (changes requested):
The code looks solid.  The tests need some refactoring.  I try (don't always succeed) to keep to TDD standards with this repo given that it is a testing library.  Nobody's going to trust a testing library with bad tests.

bitwes (approved):
Looks great, thanks.

PR #422, opened 2022-09-13
Fix "instantiate" being used as a noun or adjective

aaronfranke:
It looks like at some point there was a bulk find/replace of instance -> instantiate, which made some comments incorrect.


bitwes (approved):

bitwes:
Thank you!  This has been annoying.  I think this was done by an early version of the Godot 4 conversion tool.

PR #437, opened 2022-10-08
Recursively double nodes (closes #432)

### LINKED ISSUE #432: recursively double all nodes in a scene
So this is quite a common use case I think:

I added a script to add functionality to a parent node. Because the parent does quite some initialization and has resources that I'd rather link through the GUI, I want to make a `PackedScene` out of it and load that for my unit (and also integration) tests.

However, currently the [doubler](https://github.com/bitwes/Gut/blob/859e0a823de10c3081d1a4d9a97b94e3b82c2147/addons/gut/doubler.gd#L465) only doubles the top-level node in a scene. It would be nice to have a `recursive` option (that could default to `false`) that searches for scripts in all scene nodes and doubles them.

I understand it would make things a bit more complex, but since the [(partial) double function](https://github.com/bitwes/Gut/blob/859e0a823de10c3081d1a4d9a97b94e3b82c2147/addons/gut/test.gd#L1351-1371) already [does type-checking](https://github.com/bitwes/Gut/blob/859e0a823de10c3081d1a4d9a97b94e3b82c2147/addons/gut/test.gd#L72), maybe adding a boolean would not be too confusing?

feefladder:
Not entirely working yet (tests failing), but it has some changes to internal API, which I think would be good to put up for feedback.
Closes #432

### DIFF
--- a/addons/gut/doubler.gd
+++ b/addons/gut/doubler.gd
@@ -104,6 +104,7 @@ class ObjectInfo:
 	var _lgr = _utils.get_logger()
 	var _method_strategy = null
 	var make_partial_double = false
+	var recursive = false
 	var scene_path = null
 	var _native_class = null
 	var _native_class_name = null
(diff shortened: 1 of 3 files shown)


feefladder:
@bitwes could you have a look if this is in the right direction? then I can continue making it nicer :)

Issue #327, opened 2021-11-15
yield_to fails when code emits it with an argument


nightblade9:
**TL;DR** calling `yield(yield_to(some_thing, "some_signal", 1.0), YIELD)` fails and errors out if `some_thing` emits `some_signal` with a value (e.g. `emit_signal("some_signal", "hi there!")`).

Longer version:

I have several tests that started failing with errors similar to:

`ERROR: Error calling method from signal 'moved': 'Control(gut.gd)::_yielding_callback': Method expected 1 arguments, but called with 2..`

The simplest way to understand what's happening is that my "production code" looks like this:

```gdscript
signal moved(new_position)

func do_stuff():
  yield("moved", Vector2(37, 10))
```

In my test code, I simply use:

```gdscript
func test_stuff():
  var thing = Thing.new()
  thing.do_stuff()
  yield(yield_to(thing, "moved", 1.0), YIELD)
```

This fails with the error indicated above. I'm 99% sure it's because I started emitting `moved` with some parameter. There doesn't seem to be an easy way around this. Any help would be appreciated...

bitwes:
This was fixed in #258 and released in 7.2.0 (October 25, 2021).

nightblade9:
My bad. I even commented on that issue. Upgrading resolved my issue, thank you!

bitwes:
No problem.  I forgot when I did this and had to look up the git-blame to see when it was released.  Then I re-found 258.

jamie-pate:
This still seems broken to me?

Connects a signal to a method on a target object. Pass optional binds to the call as an [Array](https://docs.godotengine.org/en/3.3/classes/class_array.html#class-array) of parameters. These parameters will be passed to the method **after any parameter used in the call** to [emit_signal](https://docs.godotengine.org/en/3.3/classes/class_object.html?highlight=object#class-object-method-emit-signal). Use flags to set deferred or one-shot connections. See [ConnectFlags](https://docs.godotengine.org/en/3.3/classes/class_object.html?highlight=object#enum-object-connectflags) constants.

https://github.com/bitwes/Gut/blob/v7.4.1/addons/gut/gut.gd#L323 <- from_obj is the **first** argument in the callback

PR #455, opened 2023-01-20
Fix issue where yield_to can return too early

jamie-pate:
If the signal emits the value 'false' as the first argument this will assume that it's the bound 'false' value which signifies a timeout.. This is because bound arguments appear at the end of the callback parameter list, not a the beginning.

Instead of using true/false, this change binds a sentinel. We can definitively know whether the callback came from the bound signal or a timeout.

See also https://github.com/bitwes/Gut/issues/327

### DIFF
--- a/addons/gut/gut.gd
+++ b/addons/gut/gut.gd
@@ -62,6 +62,10 @@ const WAITING_MESSAGE = '/# waiting #/'
 const PAUSE_MESSAGE = '/# Pausing.  Press continue button...#/'
 const COMPLETED = 'completed'
 
+# use a class as a sentinel value since it can be used in expressions that require a const value
+class YIELD_FROM_OBJ:
+	pass
+
 var _utils = load('res://addons/gut/utils.gd').get_instance()
 var _lgr = _utils.get_logger()
 var _strutils = _utils.Strutils.new()
@@ -320,10 +324,18 @@ func _on_log_level_changed(value):
 # This is the most number of parameters GUT supports on signals.  The comment
 # on _on_watched_signal explains reasoning.
 # ------------------------------------------------------------------------------
-func _yielding_callback(from_obj=false,
+func _yielding_callback(
 		__arg1=null, __arg2=null, __arg3=null,
 		__arg4=null, __arg5=null, __arg6=null,
-		__arg7=null, __arg8=null, __arg9=null):
+		__arg7=null, __arg8=null, __arg9=null,
+		# one extra for the sentinel
+		__arg10=null):
+	var args = [
+		__arg1, __arg2, __arg3, __arg4, __arg5,
+		__arg6, __arg7, __arg8, __arg9, __arg10
+	]
+	var from_obj = YIELD_FROM_OBJ in args
+
 	_lgr.end_yield()
 	if(_yielding_to.obj):
 		_yielding_to.obj.call_deferred(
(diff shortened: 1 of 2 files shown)


addons/gut/gut.gd
```diff
@@ -61,6 +61,7 @@ const LOG_LEVEL_ALL_ASSERTS = 2
 const WAITING_MESSAGE = '/# waiting #/'
 const PAUSE_MESSAGE = '/# Pausing.  Press continue button...#/'
 const COMPLETED = 'completed'
+const YIELD_FROM_OBJ = {}
```
bitwes, line 64: Since 4.0 compares dictionaries by value this should be something else so the bug does not show back up when being ported to 4.0.  A `Node` instance would work if we make this a `var` instead...which is fine with me.  Keep the all-caps so we know it should be constant.
```
var YIELD_FROM_OBJ = Node.new()
```

addons/gut/gut.gd
```diff
@@ -320,10 +321,22 @@ func _on_log_level_changed(value):
...
+		__arg10=null):
+	var args = [
+		__arg1, __arg2, __arg3, __arg4, __arg5,
+		__arg6, __arg7, __arg8, __arg9, __arg10
+	]
+	var from_obj = false
+	for arg in args:
+		if arg == YIELD_FROM_OBJ:
+			from_obj = true
+			break
```
bitwes, line 338: Change this to:
```
var from_obj = args.contains(YIELD_FROM_OBJ)
```

test/unit/test_gut_yielding.gd
```diff
@@ -28,8 +28,13 @@ class TimedSignaler:
...
 	extends TimedSignaler
 
+	var _emitted_param
+
+	func _init(emitted_param = 1):
+		_emittied_param = emitted_param
+
 	func _on_timer_timeout():
-		emit_signal('the_signal', 1)
+		emit_signal('the_signal', emitted_param)
```
bitwes, line 37: The variable and parameter names are all wrong here. ` _on_timer_timeout` doesn't use the property, `_init` isn't assigning to the right thing.

bitwes (commented):
Good find!  I had a bit of trouble following what the issue was, but the test helped.  You have some syntax issues and one future proofing issue.  Address those and I'll take another look.

P.S.  Try running the tests before opening a PR ;).  If you are having any issues getting setup to the GUT test suite, let me know.  It's all a bit gross using the tool to test the tool.

Thanks for the PR

addons/gut/gut.gd
jamie-pate: I guess we don't care about leaking the node since gut is a singleton?

test/unit/test_gut_yielding.gd
jamie-pate: :eyeglasses:

addons/gut/gut.gd
jamie-pate: :facepalm: 

I used `arg in args`, 3.x doesn't seem to have `[].contains(value)`, it has `[].has(value)`

addons/gut/gut.gd
jamie-pate: I figured out you can use a class as a sentinel and then it can be used in places that require a constant expression. I'm not sure there is a better way to create a const sentinal with a unique identity

bitwes:
This looks good.  Thanks for the contribution!

PR #477, opened 2023-03-17
fix exit code when should_exit is set

Gerold31:
With GUT 9, gut_cmdln exits with exit code 0 even if tests failed.
This should fix this, at least if `-gexit` (`should_exit`) is set.
When manually exiting GUT (`-gexit` not set or with `-gexit_on_success`), the exit code will still be 0.
However, I don't think it's that important to have a exit code != 0, when manually running GUT,
since the exit code is mainly relevant for CI pipelines, which probably always run with `-gexit`.

Note: There is one more uncommented usage of `OS.exit_code` [here](https://github.com/bitwes/Gut/blob/godot_4/addons/gut/gut.gd#L1145).
I'm not quite sure how to fix it. Maybe just exit immediately?


bitwes (approved):
That should do it.  Thanks!

PR #480, opened 2023-04-14
Fix possible overzealous 3to4 conversion

Goutte:
`super.gutconfig.json` → `.gutconfig.json` in some feedbacks


bitwes (approved):
Good catch.

PR #487, opened 2023-05-16
Commands when $PWD contains spaces

tom-knight:
I've added double quotes around `$PWD` so that the commands work when the directory contains spaces.
I've also added bash syntax type for the markdown.


bitwes (approved):
Nice touch.

PR #499, opened 2023-06-12
chore(simulate): apply PR #409 to godot_4 branch

coffeebeats:
PR #409 has not been merged into the `godot_4` branch, so this PR applies the same changes to that branch. A few minor changes were made to the test code (in `test/unit/test_gut.gd`):

1. `assert_pass` no longer exists on `TestSimulate`, so I removed those assertions. If you'd prefer, I can update `TestSimulate` to manage assertion counts (like how `TestEverythingElse` does it) and then re-add the `assert_pass` calls.
2. I noticed that the local variable name `is_processing` in `test_simulate_calls_process_if_object_is_processing_and_check_is_true` shadows a built-in method, so I updated a number of local variable names in the tests to fix this issue while keeping naming consistent.
3. I now manually construct a few object arrays instead of using for loops (e.g. in `test_simulate_calls_process_on_descendents_if_objects_are_processing`), as they were kind of confusing when trying to understand the test.

### DIFF
--- a/addons/gut/gut_to_move.gd
+++ b/addons/gut/gut_to_move.gd
@@ -60,16 +60,34 @@ func file_touch(path):
 	FileAccess.open(path, FileAccess.WRITE)
 
 # ------------------------------------------------------------------------------
-# Call _process or _fixed_process, if they exist, on obj and all it's children
-# and their children and so and so forth.  Delta will be passed through to all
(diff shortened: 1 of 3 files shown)


bitwes (approved):
Looks good.  Thanks for porting this over.

Issue #510, opened 2023-08-15
Godot 4 some types missing from _strutils (gut/test.gd)


hilfazer:
It causes `_do_datatypes_match__fail_if_not()` to crash with an error.

Some tests where it happens:
```
extends GutTest

func test_rect2i_basis():
	assert_eq(Rect2i(), Basis())

func test_vector2i_3i():
	assert_eq(Vector2i(), Vector3i())

func test_vector4_4i():
	assert_eq(Vector4(), Vector4i())

func test_transform2d_projection():
	assert_eq(Transform2D(), Projection())

func test_callable_signal():
	assert_eq(Callable(), Signal())

```

bitwes:
should be fixed now

PR #512, opened 2023-09-19
Add some missing types

novalis:
Addresses #510 but doesn't add all of the types; these were the ones that I thought were obvious enough (and likely to come up in tests; Vector2i is what got me).


bitwes:
Thanks for the PR.  I decided to just add in all the types that were missing and add some tests to make sure things worked.  You can see the changes I made in #530.

PR #533, opened 2023-12-01
Fix untyped declaration error in Godot 4.2

andrejp88:
If the user sets the `debug/gdscript/warnings/untyped_declaration` property in their project to `Error`, tests will fail to run even if `Exclude Addons` is checked. This is probably because this particular bit of code is created dynamically from a string, so the editor doesn't realize that it's actually from an addon.


bitwes (approved):
Thanks, never would have found this.

PR #541, opened 2023-12-28
assert_eq: show raw string comparison

andymeneely:
Hello! Here's a nugget that I added to my project that saved my bacon a few times :)

In `assert_eq`, when showing the difference on a failure, long, multiline strings are hard to compare by eye. This adds an escape to the display text and lines up the expected and got so you can more easily your whitespace differences.

### DIFF
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -281,6 +281,7 @@ func assert_eq(got, expected, text=""):
 
 	if(_do_datatypes_match__fail_if_not(got, expected, text)):
 		var disp = "[" + _str(got) + "] expected to equal [" + _str(expected) + "]:  " + text
+		disp += "\nEscaped string comparison: \n" + _str(got).c_escape() + "\n" + _str(expected).c_escape()
 		var result = null
 
 		result = _compare.simple(got, expected)


bitwes:
First, thanks for the PR.  I too have been bothered by trying find the differences in large strings.  I'm not sure this is the right answer though.

I've got some issues with the code structure as it is:
*  The extra output should only be added if both things are strings (you could check for one of them to be a string since it's past the check for datatypes matching)
*  There should be a length threshold before this is used.  Then there's the question of whether or not that threshold should be configurable.  Probably, but maybe out of scope.
* `disp` should be overwritten instead of appended to (once a threshold has been put in place).  This way you don't have to see the two large strings twice.

Before you do any of that though, there might be a better approach.  We could put this in a utility function to diff strings.  It's a nice initial step.  Let's call it `diff_strings`.  You could then optionally call this if the test `is_failing()`:
``` gdscript
func test_two_big_ol_strings():
    assert_eq(big_str1, big_str2)
    if(is_failing()):
        diff_strings(big_str1, big_str2)
```
It's a few more lines of code, but it keeps output of `assert_eq` consistent, and allows anyone to do their own diff.  Though it does duplicate the output.

Thoughts?

bitwes:
Closing this, if you have any thoughts on the questions posed, reopen.

PR #544, opened 2024-01-03
Fix InputSender not emiting the `gui_input` signal

xorblo-doitus:
InputSender will now emit the `gui_input` signal. I have verified the natural order with a [minimal project](https://github.com/bitwes/Gut/files/13818986/InputOrderTest.zip), wich gave me this order:
```
_input function
gui_input signa
_gui_input function
_unhandled_input function
```

So i basically just inserted the artificial signal emission between `_input()` and `_gui_input()` artificial calls.

Bonus: I turned Strings into StringNames, because it's the desired parameter's type of `has_method()` and `has_signal()`.

### DIFF
--- a/test/unit/test_input_sender.gd
+++ b/test/unit/test_input_sender.gd
@@ -6,7 +6,11 @@ class HasInputEvents:
 
 	var input_event = null
 	var gui_event = null
+	var gui_signal_event = null
 	var unhandled_event = null
+
+	func _init() -> void:
+		gui_input.connect(_on_gui_input_emitted)
 
 	func _input(event):
 		input_event = event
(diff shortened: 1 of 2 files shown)


bitwes:
Good catch, thanks for the PR.  If you add some tests to `res://test/test_input_sender.gd` I'll merge it.

xorblo-doitus:
I added some tests and they are passing.

Note: There is an already existing test in `res://test/test_input_sender.gd` which is failing on your main branch (tested at [this point of the history](https://github.com/bitwes/Gut/tree/f5fd93f62ba60cc955a8fff7b1fa4e4955e7ed7a)), so it's not related to this pull request:
```
res://test/unit/test_input_sender.gd.TestTheBasics
- test_when_freed_all_tree_items_are_freed
    [Failed]:  Expected [@Node@26:<Node#62847452906>sender item node parent] to be freed
          at line 111
```
PS: The line of the failing test is 137 in the updated file.

test/unit/test_input_sender.gd
```diff
@@ -428,6 +454,13 @@ class TestSendEvent:
 		var event = InputEventKey.new()
 		sender.send_event(event)
 		assert_eq(r.gui_event, event)
+	
+	func test_sends_event_to_gui_input_signal():
```
bitwes, line 458: Change this test to 
```gdscript
test_send_event_causes_receiver_to_emit_gui_input_signal():
	var r = autofree(HasInputEvents.new())
	watch_signals(r)
	var sender = InputSender.new(r)
	var event = InputEventKey.new()
	sender.send_event(event)
	assert_signal_emitted_with_parameters(r, 'gui_input', [event])
```
This change will test the functionality more directly by checking for the signal and what was passed to it.

bitwes (changes requested):
This looks good, thanks, just one minor change.  

I was also thinking it would be good to verify that it does not try to emit the `gui_input` signal if it does not exist.  There really isn't a good way to test this other than using something like `pass_test('Look for engine error here')`, so I'm fine leaving that case out.

test/unit/test_input_sender.gd
xorblo-doitus: I applied your advice in [f01309b](https://github.com/bitwes/Gut/pull/544/commits/f01309bea8c2476262fb8f2e8f468b5e3f5d123d).

bitwes (approved):
Everything looks good, thanks.

PR #545, opened 2024-01-04
PoC GitHub Actions-based testing

janosdebugs:
This pull request adds a GitHub Action that tests Gut when a pull request is filed. The result will show up in the pull request and merging the pull request can be made conditional on the pull request passing.

I'm not sure if this fits your workflow, please let me know if this could be helpful.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,36 @@
+name: Test
+on:
+  pull_request:
+jobs:
+  test:
+    name: Test
+    runs-on: ${{matrix.os}}-latest
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu, windows, macos ]
+    steps:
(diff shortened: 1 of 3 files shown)


bitwes:
This could be helpful.  Right now, due to the fact that I'm using the tool to test the tool, there are tests that fail on purpose.  I've thought about doing this a little.  I'll dig into it more when I have some time.  Thanks for the contribution!

janosdebugs:
Thanks @bitwes let me play some more with this, maybe I can set up a list of expected failures and exclude those from a script.

janosdebugs:
@bitwes I've changed the pipeline to output a junit.xml file and added a script that processes the result. This would let you declare test cases that should fail. However, for some reason I can't seem to get Godot to not run in the background (daemonized), so the exit code doesn't seem to get captured correctly from the processing script. I'll keep poking at it, just posting this in case you have an idea.

Issue #506, opened 2023-07-15
Move addons/gut to its own repository and submodule


kumikumi:
As a user I want to add Gut to my project as a git submodule like this:

```
git submodule add git@github.com:bitwes/Gut.git addons/gut
```

This would allow me to use git to manage which version of Gut I have installed while not having the actual Gut files in my project repo. This would also be helpful in keeping track of which changes (if any) have been made to the Gut files (or to verify that no changes have been made to those files).

The problem here is that this repository contains other stuff besides the addon itself and the addon is not located in the root of this repo. This causes Godot to auto update the paths in import files within Gut, like this:

```
-source_file="res://addons/gut/fonts/AnonymousPro-Bold.ttf"
+source_file="res://addons/gut/addons/gut/fonts/AnonymousPro-Bold.ttf"
```

The solution would be to move the contents of `addons/gut` to a separate repository (for example `gut-plugin`), and add it to this development repo as a submodule. Projects that use Gut would only need to add the `gut-plugin` repo as a submodule instead of this full repo. This would fix the paths issue.

Releases for non-git users could be packaged from the `gut-plugin` repo.

bitwes:
I like this idea (so much that I tried it out early on) but it makes development more difficult.  I'd rather it be easier for people to contribute to GUT, it's already a bit messy to do that as it is.

The main way I handle this for my projects, is that I have GUT cloned and I've created a utility to copy the checked out version into my project.  I also add `addons/gut` to my `.gitignore`.  This makes it really easy for me to copy over the version of GUT I'm working on into a project to test changes made to GUT.  This has become a little more difficult now that there are two versions being maintained and I have 3.x and 4.x projects.  This is the standard headache of transition periods though.

In bash, I have a GUT variable and a `cpgut_here` alias that, when run from the root of a project, will copy the contents of my locally cloned GUT repo into my project.  Note that it does not clear the project's gut folder first, so you can get some extra files in there over time.
```
export GUT='/wherever/you/checked/it/out/to/Gut'
alias cpgut_here='cp -r $GUT/addons/gut/* ./addons/gut/'
```

janosdebugs:
Hey @bitwes thank you so much for your work on GUT. I'm in a similar situation as @kumikumi as I'm building several addons of my own. I'd like to avoid having to copy GUT over a dozen or so repositories.

[...]

PR #546, opened 2024-01-04
allow GUT to be placed anywhere under addons/

Dfred:
This allows GUT files to reside in an arbitrary subdirectory under addons/ with Godot 4.1+ . This simplifies filtering out addons for project export: one can set to ignore addons/tools* where non-export addons such as GUT are placed rather than maintaining a potentially long list of addon paths.

Other changes include preload to make paths more robust to change.

Sorry the linter changed the indentation.

Thanks for your tool, it proved quite handy.


janosdebugs:
@Dfred not the maintainer, but last I checked, the problem with manually editing the scene files will revert the moment you open the scene in Godot. There's a workaround for this you may be able to use, I posted the instructions here: https://github.com/bitwes/Gut/issues/506#issuecomment-1876971782

bitwes:
If you remove linter changes and can address that possible maintenance issues reopen.

PR #575, opened 2024-02-26
add debugger version of gut_cmdln

bitwes:


smks:
Thanks for continuing to support Godot v3 :)

PR #577, opened 2024-03-11
Wait until object is freed

WebF0x:
I think it could be useful. For example, in my game I want to test that when I hit an enemy, then it dies and is freed. But I found it awkward to await on signals on this soon-to-be null object

# Before

```
enemy.set_name('node_name')
enemy.set_linear_velocity(Vector2.LEFT * 1000)

# Have to wait, but it's hard to guess for how long and slows test runs
wait_seconds(10)
# Or create a custom signal just for the test. It works but gives this awkward error message ==> ERROR: Parameter "obj" is null.
await wait_for_signal(enemy.damaged, 10)

assert_freed(enemy, 'node_name')
```

# After

```
enemy.set_name('node_name')
enemy.set_linear_velocity(Vector2.LEFT * 1000)

await wait_until_freed(enemy, 10)

assert_freed(enemy, 'node_name')
```

### DIFF
--- a/addons/gut/awaiter.gd
+++ b/addons/gut/awaiter.gd
@@ -7,6 +7,8 @@ var _wait_time = 0.0
 var _wait_frames = 0
 var _signal_to_wait_on = null
 
+var _object_waiting_to_be_freed: Variant = null
+
 var _elapsed_time = 0.0
 var _elapsed_frames = 0
 
@@ -22,6 +24,10 @@ func _physics_process(delta):
 		if(_elapsed_frames >= _wait_frames):
 			_end_wait()
 
+	var is_waiting_for_object_to_be_freed := hash(_object_waiting_to_be_freed) != hash(null)
+	if is_waiting_for_object_to_be_freed:
+		if not is_instance_valid(_object_waiting_to_be_freed):
+			_end_wait()
 
 func _end_wait():
 	if(_signal_to_wait_on != null and _signal_to_wait_on.is_connected(_signal_callback)):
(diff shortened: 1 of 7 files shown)


addons/gut/awaiter.gd
```diff
@@ -22,6 +24,10 @@ func _physics_process(delta):
 		if(_elapsed_frames >= _wait_frames):
 			_end_wait()
 
+	var is_waiting_for_object_to_be_freed := hash(_object_waiting_to_be_freed) != hash(null)
```
bitwes, line 27: Why the use of `hash` when checking for `null`?

bitwes (changes requested):
This is a great PR.  Thank you.  I was a little on the fence as to whether this feature was actually needed, but you've done ALL the work so I think this will make a nice addition.  I have one question about the use of `hash`.  Once that is squared away I'll be good to merge this.

WebF0x:
Thanks! Hey let's hold on for this PR until we resolve https://github.com/bitwes/Gut/issues/585, as it might be a more generic solution.
I don't have permission for it, but you can mark it as a "draft PR"

addons/gut/awaiter.gd
WebF0x: With hindsight it does look weird. I probably hit a case where I needed to differentiate null and freed objects like mentioned [in this forum post](https://forum.godotengine.org/t/distinguish-between-freed-object-and-null/3838/2).

If `_object_waiting_to_be_freed != null` still passes all tests, it would sound simpler. Otherwise documenting why it has to be weird will help future readers!

WebF0x:
Closing in favor of: https://github.com/bitwes/Gut/pull/609

PR #587, opened 2024-04-03
WIP: Bugfix/i578

### LINKED ISSUE #578: Input does not seem to receive events from an InputSender
I'm not sure if the error is on my side or if there is a bug. When i try to test my scene that uses `Input` to query events in the `_physics_process ` function, it always returns false or in the case of `Input.get_vector` a zero vector. I did follow the instructions found on the wiki.

If i missed any information that would help, please let me know.

OS: Linux
Godot version: 4.2.1
Gut version: 9.2.1
Example project: [test.zip](https://github.com/bitwes/Gut/files/14563474/test.zip)

lxkarp:
The hope here is to resolve #578

I haven't done any cleanup as suggested by @bitwes over in the issue thread. But the fix he suggested works just fine, and this PR is evidence of that.

### DIFF
--- a/addons/gut/input_sender.gd
+++ b/addons/gut/input_sender.gd
@@ -277,6 +277,11 @@ func _send_event(event):
 	for r in _receivers:
 		if(r == Input):
 			Input.parse_input_event(event)
+			if(event is InputEventAction):
+				if(event.pressed):
+					Input.action_press(event.action)
+				else:
+					Input.action_release(event.action)
 			if(_auto_flush_input):
 				Input.flush_buffered_events()
 		else:
--- a/test/unit/test_bugs/test_i578.gd
+++ b/test/unit/test_bugs/test_i578.gd
@@ -1,0 +1,43 @@
+extends GutTest
+
+class InputSingletonTracker:
+	extends Node
+	var input_frames = []
+
+	var _frame_counter = 0
+
+	func _process(delta):
+		_frame_counter += 1
+		if Input.is_anything_pressed():
+			input_frames.append(_frame_counter)
+
+class TestInputSingleton:
+	extends "res://addons/gut/test.gd"
+	var _sender = InputSender.new(Input)
+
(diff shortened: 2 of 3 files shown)


lxkarp:
I think this may not correctly trigger `is_action_just_pressed`. I'll make another test to validate this tomorrow.

Kulkodar:
Thank you for taking the initiative, with the changes all my tests are working again. It seems that ´is_action_just_pressed´ is working.

lxkarp:
I found that the problem with `is_action_just_pressed` -- for me at least -- is that it doesn't work in `_process` but it does work in `_physics_process`

bitwes (changes requested):
Thanks for the PR.  The code change looks good.  I'd like some additional testing to verify that the `Input.is_action_just_*` methods work as expected.  If you change the `InputSingletonTracker` to the code below, you should be able to make assertions that they are working.

**NOTE** When I tried verifying that `is_action_just_pressed` was working I had to add a leading `wait_frames` for the `InputSingletonTracker` to detect the frame when the action goes down
``` gdscript
_sender.wait_frames(1).action_down("jump").hold_for('10f')
await wait_for_signal(_sender.idle, 5)
```

Here's my take on `InputSingletonTracker`
```gdscript
class InputSingletonTracker:
	extends Node
	var pressed_frames = []
	var just_pressed_count = 0
	var just_released_count = 0

	var _frame_counter = 0

	func _process(delta):
		_frame_counter += 1

		if(Input.is_action_just_pressed("jump")):
			just_pressed_count += 1

		if(Input.is_action_just_released("jump")):
			just_released_count += 1

		if Input.is_action_pressed("jump"):
			pressed_frames.append(_frame_counter)

```

bitwes:
Hey @lxkarp.  @Edearth added the requested tests so I'm merging this PR for street cred, then I'll merge theirs.  Thanks!

PR #600, opened 2024-05-09
Automatically enable `exclude_addons` option when running tests

mphe:
GUT will throw errors if `exclude_addons` option is disabled, e.g. due to overriding native methods.
When working on addons it is usually desired to disable `exclude_addons` in order to get proper code diagnostics.
This however makes it tedious to run GUT because the option has to be manually disabled before running tests and manually enabled again afterwards to get diagnostics.
To workaround this issue, this commit automatically enables the `exclude_addons` option when running tests, so regardless of the project settings in the editor, GUT will always run as intended.

### DIFF
--- a/addons/gut/gut_config.gd
+++ b/addons/gut/gut_config.gd
@@ -56,6 +56,17 @@ var options = default_options.duplicate()
 var json = JSON.new()
 
 
+func _init() -> void:
+	# Ensure the exclude_addons option in the project settings is enabled when running GUT.
+	# This is important, because GUT will throw errors if this option is disabled, e.g. due to
+	# overriding native methods.
+	# When working on addons it is usually desired to disable exclude_addons in order to get proper
+	# code diagnostics.
+	# This however makes it tedious to run GUT because the option has to be manually disabled before
+	# running tests.
+	if not Engine.is_editor_hint():
+		ProjectSettings.set("debug/gdscript/warnings/exclude_addons", true)
+
 func _null_copy(h):
 	var new_hash = {}
 	for key in h:


bitwes (changes requested):
This is a great idea.  I double checked, and this cannot be done in a pre-hook script.  This must be done before an instance of `addons/gut/utils.gd` is created, since it loads all the scripts that have issues.  

I think a better place for this (I tested a bunch of different spots) is a new `_init()` method in `addons/gut/gut_config.gd`.  Although I cannot think of a scenario where you would not want to enable this, since it is on by default and it is suggested that you enable it if it is not, there still might be a reason.  If that ever arises, having it in `gut_config.gd` will make more sense when an option is added.

mphe:
Isn't gut_config also instantiated in the editor? In that case it should not change `exclude_addons` because it would interfer with the user's project config.
The intent was to only change it at runtime.

But I sure can add a setting to enable or disable it if needed.

bitwes:
Good catch.  I just tested it, and it will turn it back on when you open the project.  I added a check for `is_editor_hint` and it does not turn it back on then.

``` gdscript
func _init():
	if(!Engine.is_editor_hint()):
		ProjectSettings.set("debug/gdscript/warnings/exclude_addons", true)
```

I don't think an option is needed yet.  I'm not entirely sure where it would be applied, mostly because "Native Method Override" is an error by default (which I have no issue with) .  I don't know if there is a point in execution where we have loaded a config file and `utils.gd` has NOT been referenced (which loads the problem scripts).  We may only be able to turn "exclude addons" off at a later time, instead of having it always off based on the option.  

I tried lazy loading the problem scripts but that didn't stop the errors when opening the project.  I'm not sure if this would help in finding a spot to apply an option for auto-exclude-addons.
``` gdscript
var _Doubler = null
var Doubler = null :
	get:
		if(_Doubler == null):
			_Doubler = load('res://addons/gut/doubler.gd')
		return _Doubler
```

mphe:
Moved the code into `_init()` in `gut_config.gd`.

bitwes (approved):
Thanks

PR #610, opened 2024-05-22
Manage Warnings and static `utils.gd`

bitwes:
This PR aimed to manage warnings for GUT scripts and user test scripts.  Early on it became clear that it would benefit greatly from converting `GutUtils` to a completely static class.  This was a little bit of feature creep but was high on the todo list.

Making `GutUtils` static allowed for the lazy loading of GUT scripts which in turn allowed us to control when the scripts were loaded.  Loading the scripts at a specific time allows us to alter warning levels before loading them.  This means that users that have more strict warnings, or do not want to exclude addons, will not see warnings generated from GUT scripts.

This PR introduces the following new internal features:
* `res://addons/gut/lazy_loader.gd`:  This script is used to load scripts when they are first referenced or the static method `load_all` is called.  In conjunction with how it is used in `GutUtils`, we can defer loading scripts.
* `res://addons/gut/warnings_manager.gd`:  This static script has tools to simplify changing Project Setting warnings.  This is used by GUT at key times to change warning levels.
* `GutUtils`:  
  * Completely static, no more pseudo global/singleton instance.
  * All references to `addons/gut/utils.gd` have been replaced with `GutUtils`
  * All `_utils` variables have been removed and `GutUtils` is used instead.
  * Definitions for warnings settings to be used at key times while running test scripts have been added.
* Some functionality has been moved from `GutUtils` to their own scripts to simplify the static logic in `GutUtils`.
  * `addons/gut/version_numbers.gd`
  * `addons/gut/dynamic_gdscript.gd`

Warning changes during run:
*  The warnings setting `exclude_addons` is enabled at the start of the run, and remains enabled for the durtation of the run.
*  Warnings are disabled (`enable = false`) before loading each test script is loaded.  If warnings were enabled before loading the script, they are re-enabled after.


bitwes:
@mphe Would you mind giving this PR a test spin?   This PR re-addresses the disabling of `exclude_addons` at the start of a run, and it also disables all warnings when loading `GutTest` scripts.  It also implements functionality that will enable more fine grained warning management going forward.  I'm planning to add a hook script right after all GUT scripts get loaded so the user can set their own warnings if needed.

Also, your idea to put warning management code in `GutRunner._ready` ended up being really good.  With the lazy loading logic implemented in this PR, we defer the loading of most GUT scripts until that point, and it makes things much cleaner.  Although this PR does undo some of your changes, it's completely inspired by them.   Thanks for all your feedback and PRs.

mphe:
This sounds pretty good! 
I'm currently not at home and won't be back until the weekend. I'll report back in the following week.

mphe:
Anything specific I should watch out for while testing this PR?
Been testing recent master and everythings works so far as usual.

bitwes:
I merged it into main 2 days ago, so if you've been using that then everything should be good.  I've disabled exclude addons in GUT and have been dog flooding the change without issue.

PR #621, opened 2024-06-09
Fix #618

### LINKED ISSUE #618: assert_string_contains does not check datatype prior to performing string operations
# Versions
* Godot: 4.2
* GUT: 9.2.1 (main)
* OS: macOS Sonoma

# The Bug
```
func test_something():
    assert_string_contains(null, 'this breaks', false)
```
That test will cause the error:  
```
Debugger Break, Reason: 'Invalid call. Nonexistent function 'to_lower' in base 'Nil'.'
*Frame 0 - res://addons/gut/test.gd:849 in function 'assert_string_contains'
```

This should generate a failure that GUT cannot convert nil to String, but it breaks isntead.

nycdotnet:
This PR attempts to fix issue #618 .  This appears to fix the original issue, but there are two concerns I have that I need your opinion on.

Concerns:
1. The message is not entirely clear in the case of when someone passes a number - in fact it looks like it should have worked because of the formatting (see below).  Should I go through the trouble of improving the formatting?  There's a nice method here from user Jengamon that would be useful but I don't know if this project can just incorporate code like that, or maybe if something already exists here.  https://forum.godotengine.org/t/is-there-a-corresponding-type-function-in-gdscript/31815/4
2. Should we automatically convert `text` to a string if it's not null?  Maybe that's more aligned with GDScript as a dynamically typed language? I lean towards no, but I can see the argument that testing if "2" appears in the number `123` should be true.  I am more strongly against doing it the other way (testing if the number 2 appears in the string "123") because that feels more like a defective test.

Example errors which I think are fine:
```
Expected text and search to both be strings.  You passed '<null>' and 'whatever'.
Expected text and search to both be strings.  You passed 'hello' and '<null>'.
```

Example errors with which I have concerns:
```
Expected text and search to both be strings.  You passed '123' and '2'.
Expected text and search to both be strings.  You passed '2' and '123'.
```

### DIFF
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -836,9 +836,12 @@ func assert_not_typeof(object, type, text=''):
 # The match_case flag determines case sensitivity.
 # ------------------------------------------------------------------------------
 func assert_string_contains(text, search, match_case=true):
-	var empty_search = 'Expected text and search strings to be non-empty. You passed \'%s\' and \'%s\'.'
+	const empty_search = 'Expected text and search strings to be non-empty. You passed \'%s\' and \'%s\'.'
+	const non_strings = 'Expected text and search to both be strings.  You passed \'%s\' and \'%s\'.'
 	var disp = 'Expected \'%s\' to contain \'%s\', match_case=%s' % [text, search, match_case]
-	if(text == '' or search == ''):
+	if(typeof(text) != TYPE_STRING or typeof(search) != TYPE_STRING):
+		_fail(non_strings % [text, search])
+	elif(text == '' or search == ''):
 		_fail(empty_search % [text, search])
 	elif(match_case):
 		if(text.find(search) == -1):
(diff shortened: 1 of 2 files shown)


bitwes:
> The message is not entirely clear in the case of when someone passes a number

You can use test.gd's `_str` method (which is a legacy wrapper around `strutils.gd type2str`) to do some formatting.  I think it will turn strings into "string" and not do any formatting to numbers (so you can remove the single quotes from the strings you are making).

You can also use `_strutils.types[typeof(search)]` to get a string representation of the type constant (which is GUT's version of that method you linked to).

> Should we automatically convert text to a string if it's not null?

No.  I've had to fight urges to do stuff like this in the past.  This could cause false positives if the search term is the result of code under test that should be returning a string but is not.  It's best to make the user explicitly convert values.  This is why many asserts use `_do_datatypes_match__fail_if_not` (which doesn't really apply to this since they are both supposed to be strings).

bitwes (approved):
This looks great.  Thanks for doing the TODO as well.

Issue #624, opened 2024-06-10
Add `assert_gte` and `assert_lte`


nycdotnet:
# Versions
What versions of Godot do you want to use this feature in?

Godot 4.x, but would be appropriate for any supported version.

# The Feature

GUT currently supports `assert_lt` for less than and `assert_gt` for greater than, but not `assert_lte` for less than or equal to, or `assert_gte` for greater than or equal to.  Is there an affirmative reason for this?  If not, I would love to add this.  Thanks.

bitwes:
I want to say this had been discussed in the past, but I couldn't find anything in the issues.  I don't have anything against creating the asserts.  Yea, this would be a nice bit of polish.  Let me know if you have any questions about implementing these.

nycdotnet:
This is done with #625 .

PR #625, opened 2024-06-11
Implement `assert_lte` and `assert_gte`

nycdotnet:
This implements #624 .

### DIFF
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -383,12 +383,34 @@ func assert_gt(got, expected, text=""):
 			_fail(disp)
 
 # ------------------------------------------------------------------------------
+# Asserts got is greater than or equal to expected
+# ------------------------------------------------------------------------------
+func assert_gte(got, expected, text=""):
+	var disp = "[" + _str(got) + "] expected to be >= than [" + _str(expected) + "]:  " + text
+	if(_do_datatypes_match__fail_if_not(got, expected, text)):
+		if(got >= expected):
+			_pass(disp)
+		else:
+			_fail(disp)
+
+# ------------------------------------------------------------------------------
 # Asserts got is less than expected
 # ------------------------------------------------------------------------------
 func assert_lt(got, expected, text=""):
 	var disp = "[" + _str(got) + "] expected to be < than [" + _str(expected) + "]:  " + text
 	if(_do_datatypes_match__fail_if_not(got, expected, text)):
 		if(got < expected):
+			_pass(disp)
+		else:
+			_fail(disp)
+
+# ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


nycdotnet:
I can write up the docs too if you're satisfied with the implementation.

bitwes:
Implementation looks good.  Adding to the docs would be great.

I had a thought about the tests.  The tests you have are fine, but I think these should all be done as a single parameterized test (which came along well after most of the asserts were made and tested).  I was thinking about adding a few tests just to see how it behaved.  It would be a lot easier to do that with parameterized tests.  

I did the gte tests.  First I added these constants to the `BaseTestClass` at the top of the script (`test_test.gd`)
```gdscript
	const SHOULD_PASS = &'Should Pass'
	const SHOULD_FAIL = &'Should Fail'
```
Then I implemented the tests you have as
``` gdscript
	var _basic_values = ParameterFactory.named_parameters(
	['v1', 'v2', 'pass_fail'],
	[
		[2, 1, SHOULD_PASS],
		[1, 2, SHOULD_FAIL],
		[10, 10, SHOULD_PASS],
		['a', 'b', SHOULD_FAIL],
		['b', 'a', SHOULD_PASS],
		['A', 'A', SHOULD_PASS]
	])
	func test_basic_values(p = use_parameters(_basic_values)):
		gr.test.assert_gte(p.v1, p.v2, p.pass_fail)
		if(p.pass_fail == SHOULD_PASS):
			assert_pass(gr.test, 1, str(p.v1, ' >= ', p.v2))
		else:
			assert_fail(gr.test, 1, str(p.v1, ' >= ', p.v2))
```

I think this is better, and would like to do this going forward, but not better enough to make someone change it.  If you think it's better and would like to use this method, then go ahead and change `TestAssertGte` and `TestAssertLte`.  Not a big deal if you don't want to.

nycdotnet:
Is this what you had in mind or should I combine them into one?

nycdotnet:
Docs implemented - this is ready for review.

bitwes (changes requested):
The documentation and the tests themselves look good.  I'd like to see the tests put back into their own individual inner classes (`TestAssertLt`, `TestAssertLte`, `TestAssertGt`, `TestAssertGte`).  The test cases are all very similar, but I prefer my tests to be more DAMP than DRY (I just found out that [DAMP](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) was a thing; aka MOIST aka AHA aka WET.  I was going to say "not DRY", but of course we named it.)

nycdotnet:
ok.  Love "DAMP" - will use that in future.

bitwes (approved):
Perfect.  Thanks for all your work.

PR #628, opened 2024-06-18
Assert eventually refactor

bitwes:
* Refactors the `awaiter` out of `gut.gd` and into `test.gd` which greatly simplifies awaiting.  This was a holdover from the way `yields` are handled in 3.x.
* Adds `wait_until` to `test.gd`.  This returns `true` if it does not timeout, false otherwise.
* `wait_for_signal` now returns `true` if the signal was emitted before timeout, `false` otherwise.  This means you can do `assert_true(await wait_for_signal(...`.
* Removes `assert_eventually`, preferring the more generic approach of `asssert_true(await wait_until(...`.
* Adds ability to wait an amount of time between calling predicate methods to handle cases where the method is expensive or any other reason you can think of.
* `wait_until` (in awaiter.gd and therefore test.gd) ignores non-boolean values, and only ends the wait for `true`.


bitwes:
@WebF0x I've spent way too much time thinking about `assert_eventually` and predicate functions and names and usage.  I've come up with a solution I like a little better than the initial implementation.  I'd like to get your take on this approach and see if you can poke any holes in it.  

The main points:
* added `wait_until` wrapper to `GutTest`.
  * This returns `true` if the predicate method returns `true` (explicitly) before timing out, false otherwise.
* Altered `wait_for_signal` to return `true` if signal is emitted before timing out.
* These changes allow for a general approach to awaiting and asserting:
```
assert_true(await wait_for_signal(my_obj.my_signal, 10),
    "Signal should emit before we hit 10 seconds")
assert_true(await wait_for(my_method, 10),
    "Method should return true before 10 seconds")
```

* removed `assert_eventually` in lieu of `assert_true(await wait_for(predicate, ...`.  The new approach is a bit more verbose than just calling `assert_eventually`, but it allows for a more general pattern that can be used with `wait_for_signal`.  I think `assert_eventually` or some variant of it will likely be reintroduced once usage patterns are a lot more clear.

test/unit/test_test_await_methods.gd
```diff
@@ -80,79 +87,86 @@ class TestTheNewWaitMethods:
...
+		var method = pred_methods.called_x_times.bind(10)
+
+		await wait_until(method, 1.1, 'DID YOU SEE THIS?')
+		pass_test("Check output for DID YOU SEE THIS?")
+
+	func test_wait_until_accepts_time_between():
+		var pred_methods = PredicateMethods.new()
+		var method = pred_methods.called_x_times.bind(10)
+
+		await wait_until(method, 1.1, .25)
```
WebF0x, line 153: Why have the 3rd param be either msg_or_time_between?
At first glance I found it a bit confusing.

I wish GDScript had [named parameters](https://github.com/godotengine/godot/issues/6866#issue-183827265) for stuff like this
Maybe pass in a Dict parameter for all the "[options](https://docs.godotengine.org/en/stable/classes/class_dictionary.html)" (timeout_s, time_between_s, message, etc)
```
wait_until({"timeout_s": 1.1, "time_between_s": .25, "message": "bleerg"})
```
Then any option could be omitted and use defaults values.

Not a big deal, just food for though

test/unit/test_test_await_methods.gd
```diff
@@ -80,79 +87,86 @@ class TestTheNewWaitMethods:
...
+		await wait_until(all_is_good, .5)
+		assert_almost_eq(counter.time, .5, .05)
+		assert_true(did_wait_timeout(), 'did_wait_timeout')
+
+	func test_wait_until_returns_true_when_it_finishes():
+		var all_is_good = func():
+			return counter.time > .25
+
+		var result = await wait_until(all_is_good, .5)
+		assert_true(result)
```
WebF0x, line 133: Pros:
- It opens the door to testing that "something won't happen"
  ```
  hero.strike_softly()
  assert_false(await wait_until(is_dead, 1))
  ````
- `wait_until` more coherent with the other `wait_*` functions

Cons:
- Easy to forget the assertion and mess up the test
  ```
  hero.strike_strongly() # Oops actually a soft hit
  await wait_until(is_dead, 1) # Oops it's not dead but still passes
  ```
- `wait_until(is_dead)` does not make it obvious if true or false is a "success". Maybe we could rephrase it as something like `eventually(is_dead)` which reads more easily in my mind
  - Btw English is my 2nd language, you can probably come up with a more natural phrasing haha

So I'm thinking the ideal world would be an `assert_eventually()` for most cases, and `eventually()` for rarer advanced use cases.
But it's really just syntax sugar at this point, nothing critical.

WebF0x (commented):
I like that your way opens new doors for expanding the idea in the future.
Here's are my first impressions (no major holes)

test/unit/test_test_await_methods.gd
bitwes: Named parameters would be great.  There's a couple places in GUT where I've had to take this approach and I've never been happy with it.  It's always ugly trying to pick which should be the last optional parameter and what they should be named.  Maybe just `p3` and `p4` would be less confusing?  They don't tell you anything but they aren't misleading either.

Honestly, the time-between feature smells of feature creep.  I talked myself into it because it seems more difficult to implement later.  And if I did add it later, the method signature would end up looking this way so it was backwards compatible (a horrible justification, I know).

I've thought about dictionaries for optional parameters.  I think they are a bad user experience though, they are annoying to type and more vulnerable to typos.  On the implementation side you have to validate the dictionary before you do anything.  GUT can cause a test to fail with an error, but you don't want the end user to have to check return values just for valid parameters.  It's all stuff that can be worked around, but its never felt worth it for a few rare cases.

Maybe it should just be two methods:  `wait_until`, `wait_until_with_time_between` (it's such a long name though).

Naming things is the worst.

test/unit/test_test_await_methods.gd
bitwes: > Easy to forget the assertion and mess up the test

GUT will mark any test that does not assert as risky so you know you didn't assert.  Also, when writing tests you should see them fail once before you implement the code that would make them pass.  Otherwise you can't know if you have tested anything (Red/Green/Refactor).

> wait_until(is_dead) does not make it obvious if true or false is a "success". 

The name of the test should give the `wait_until` the context it needs.  In this example I would expect the name to be something like `test_strong_strike_kills_player`.

We could add `wait_while` to do the inverse of `wait_until`.  This would wait until the predicate does not return `true`.  Then you could do something like `wait_while(is_alive)`.

WebF0x: > The name of the test should give the wait_until the context it needs. In this example I would expect the name to be something like test_strong_strike_kills_player.

What I meant is that when you read"wait_until" it's not clear what the returned true or false mean. You have to read the docs (different libs use true as success and others false as success). So if the method name made it obvious it would be a plus. (Think is_alive => it's obvious that true means that the thing is alive, but if it was named check_alive, now you have to guess)

WebF0x: > We could add wait_while to do the inverse of wait_until

Good idea!

If we had these 2, I can't think of a reason to wait until/while something and not assert on the result.
And if we always assert on the result, might as well re-add the assertion inside the wait_* methods.

WebF0x: > GUT will mark any test that does not assert as risky so you know you didn't assert. Also, when writing tests you should see them fail once before you implement the code that would make them pass. Otherwise you can't know if you have tested anything (Red/Green/Refactor).

Totally agree and I always try to do that as much as possible. But that's an ideal world. The real world is messy, tests might have multiple assertions and GUT users will have varying levels of TDD skills

bitwes: > What I meant is that when you read"wait_until" it's not clear what the returned true or false mean

I understand now.  I had to take a step back and reassess, and it is a bit ambiguous.  My thought is that `wait_until(is_dead, x)` is returning what the predicate returned on the last call.   You would ignore "wait_until" when reading the line as English.  `if(await wait_until(is_dead, 10))` would be "If (wait a bit) is dead then".

Ok, ok, ok...you are onto something.  "wait a bit" is the same thing as "eventually".  I'm going to think on that a bit.  I may leave it as it is for now.  It would be easy to make syntax sugar or rename it and deprecate the old name.

> Good idea!

Cool.  I'm going to make an issue for `wait_while` so that this PR doesn't keep growing.

> If we had these 2, I can't think of a reason to wait until/while something and not assert on the result.

There still might be a reason to use these without asserts in integration tests.  For example, if you have a cooldown for attacks, you might use these to strike multiple times without having to worry that the script might hang forever.
```
player.hit(enemy)
wait_until(player.can_attack, 10)
player.hit(enemy)
wait_until(player.can_attack, 10)
player.hit(enemy)
assert_true(await wait_until(enemy.is_dead, 10))
```
Here, 10 is just some arbitrarily large value that you never expect to hit, not a time limit you are testing against.

> The real world is messy, tests might have multiple assertions and GUT users will have varying levels of TDD skills

I agree, it is really tough deciding scope.  In the end GUT is a TDD tool, so I have to treat it that way or I'll never stop tweaking things.  I really like tweaking things.

WebF0x: When the behavior is not what you expected, wouldn't you want it to fail early?

It would give you more information
- The test did not fail because 3 hits failed to kill the enemy
- The test failed because the player was only able to hit 1 time when I wanted it to hit 3 times

```
player.hit(enemy)
wait_until(player.can_attack, 10) # Timeouts, dang it!
player.hit(enemy)
wait_until(player.can_attack, 10)
player.hit(enemy)
assert_true(await wait_until(enemy.is_dead, 10))
```

bitwes: This would be after you have tested the more simple interactions and you can assume that functions are working properly based on previous tests.  Probably an integration test.

bitwes:
Based on all the conversations, this is the list of the outstanding questions I/you had.  I'm going to merge this and move on.  If anything comes up we'll make an issue.

* Should wait_until be error if the predicate returns a non-boolean?  
  * Since it is checking exclusively for `true` in order to stop, I don't think it is a big deal to treat anything else as "not true". 
* Created issue for `wait_while` (#630)
* I renamed the confusing parameters to `p3`, `p4` and updated the documentation.
* `eventually`/`assert_eventually` will be addressed later.  I think names and uses will become a little more clear as this feature gets used more.

Issue #657, opened 2024-08-30
On small resolutions the runner takes more space than available


Edearth:
# Versions

* Godot: 4.3.1
* GUT: 9.2.1
* OS: Linux

# The Bug

On small resolutions the runner takes more space than available in the screen. As a result, it appears cropped like this:
![image](https://github.com/user-attachments/assets/c91558be-ff05-405f-8f3c-0065d786e258)

I'm working on a game with a resolution of 384x216. I would expect the window gets resized correctly and I can see all of it. Like this:
![image](https://github.com/user-attachments/assets/2e5e2d32-ee4a-4c9b-beb3-af96249ca0f6)

# Steps To Reproduce

1. In Godot, create a new project and a test
2. In Godot, open `Project > Project Settings > Display > Window`
3. Set `Viewport Width` to 384 and `Viewport Height` to 216.
4. (Optionally, in GUT's settings, make sure `Exit on Finish` and `Exit on Success` are disabled)

# Additional notes

Maybe there could be an option for it in GUT's settings. Setting a custom size, or a checkbox to make it fullscreen.

bitwes:
I could probably decrease `custom_minimum_size.x` on the normal gui (or add the min size logic from the script below).  That doesn't really help you out that much though, since it would still be off screen.  Maybe exposing the NormalGui and CompactGui nodes to `GutHookScript` would be the most versatile approach.  This would allow you to do whatever you wanted to the gui nodes in a [pre-hook script](https://gut.readthedocs.io/en/latest/Hooks.html).

Here's a pre-run hook script that accesses some "private" variables to get to the normal gui and then alter it to fit on the screen.  Try this out, and let me know if it solves the issue for you.  If so, then making it easier to get to these gui nodes in a hook script is probably the best approach.

I usually put my hook scripts in `res://test/resources`.  You need to add this in your `.gutconfig` and/or configure it in the editor as the pre-hook script.

<img width="387" alt="image" src="https://github.com/user-attachments/assets/040bd119-f1fd-4279-bc4a-d128d6200c0d">

``` gdscript
extends GutHookScript

func run():
	var runner = gut.get_node("/root/GutRunner")
	var win_size = gut.get_viewport().get_visible_rect().size
	var norm_gui = runner._gui._normal_gui
	var min_size = Vector2(
		min(win_size.x, norm_gui.custom_minimum_size.x),
		min(win_size.y, norm_gui.custom_minimum_size.y)
	)
[...]

PR #658, opened 2024-08-30
fix: make runner take all screen, remove min size

Edearth:
This is a simple fix for #657. It forces the runner to take almost all available space (except a margin of 5 pixels on each side). The result looks like this on a project with a resolution of 384x216:

image

But maybe you don't want this to be the default behavior. I was thinking it could also be an option in GUT's settings. Like setting a custom size, or a checkbox to make it fullscreen.

### DIFF
--- a/addons/gut/gui/NormalGui.tscn
+++ b/addons/gut/gui/NormalGui.tscn
@@ -6,7 +6,6 @@
 [ext_resource type="PackedScene" uid="uid://bvrqqgjpyouse" path="res://addons/gut/gui/ResizeHandle.tscn" id="4_2r8a8"]
 
 [node name="Large" type="Panel"]
-custom_minimum_size = Vector2(500, 150)
 offset_right = 632.0
 offset_bottom = 260.0
 theme = ExtResource("1_5hlsm")
--- a/addons/gut/gui/gut_gui.gd
+++ b/addons/gut/gui/gut_gui.gd
@@ -227,6 +227,7 @@ func to_bottom_right():
 
 func align_right():
 	var win_size = get_display_size()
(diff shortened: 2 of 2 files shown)


bitwes:
I could probably decrease `custom_minimum_size.x` on the normal gui (or add the min size logic from the script below).  That doesn't really help you out that much though, since it would still be off screen.  Maybe exposing the NormalGui and CompactGui nodes to `GutHookScript` would be the most versatile approach.  This would allow you to do whatever you wanted to the gui nodes in a [pre-hook script](https://gut.readthedocs.io/en/latest/Hooks.html).

Here's a pre-run hook script that accesses some "private" variables to get to the normal gui and then alter it to fit on the screen.  Try this out, and let me know if it solves the issue for you.  If so, then making it easier to get to these gui nodes in a hook script is probably the best approach.

I usually put my hook scripts in `res://test/resources`.  You need to add this in your `.gutconfig` and/or configure it in the editor as the pre-hook script.

image

``` gdscript
extends GutHookScript

func run():
	var runner = gut.get_node("/root/GutRunner")
	var win_size = gut.get_viewport().get_visible_rect().size
	var norm_gui = runner._gui._normal_gui
	var min_size = Vector2(
		min(win_size.x, norm_gui.custom_minimum_size.x),
		min(win_size.y, norm_gui.custom_minimum_size.y)
	)
	norm_gui.custom_minimum_size = min_size
	norm_gui.size = min_size
	norm_gui.align_right()
```

Edearth:
Closing since it can be fixed with a [hook script](https://github.com/bitwes/Gut/issues/657#issuecomment-2322914129).

PR #663, opened 2024-09-14
Fixed typos in error messages.

xmoby:
I was hit with the "configrued" error message, so I decided to go and review the other messages. I found three small things to fix.


bitwes (approved):
Thank you.  Parameter is one of my favorites to mess up.  I kinda like "configrued" as a word, but you are right, it should be changed.

Edit:  "mess of" -> "mess up"...eh.

Issue #670, opened 2024-11-27
Await in Pre-Run Hook Script


LowFire:
# Versions
What versions of Godot do you want to use this feature in?
4.x

# The Feature
I would like the ability to await in the pre-run hook script, meaning, you can halt the execution of the tests until something happens. I don't know if this is already possible or not, but from what I could tell, it doesn't seem like you can do this. Correct me if i'm wrong.

So here's a use case example of why I think this could be useful, and it applies to the game project I'm currently working on. My game is multiplayer, so some of my tests require a multiplayer connection. I've created a test server that responds to requests from a test client. The client runs the tests. It sends requests to the server. The server receives the requests and does stuff that's related to the test that the client is currently running. Server sends results back to the client. Client runs asserts on the results. That's the short of what i'm doing.

I try to spin up my test server in the pre-run hook script, but here's the problem. I don't want to run my tests until a connection has been established with the test server. This is to prevent the possibility that a test is ran that requires a multiplayer connection before a connection has even been established. For that reason, I want to be able to pause before the tests are ran so that there is time for a connection to be made. Additionally, if for whatever reason, connection to the test server has failed, I want to be able to abort the tests using `abort()`

LowFire:
I dug into GUT's source code and hacked it in. Here's what I did.

In gut.gd:

Added an awaiter property in gut.
![gut1](https://github.com/user-attachments/assets/e7b458f6-0e1e-4de8-8b72-729d82102d1c)

Add the awatier as a child of gut in the ready function.
![gut3PNG](https://github.com/user-attachments/assets/fcdad75f-6212-43bd-a17d-92b82a66cc7e)

Then I added the await keyword before `_run_hook_script` on the pre-run script.
![gut2](https://github.com/user-attachments/assets/bc21504e-3763-4ec4-bbb8-e652b9931f79)

And this is what my pre-run hook script looks like:
![gut4](https://github.com/user-attachments/assets/2e4d6f63-c5f2-49ce-87e3-db0ebb2173bb)

So basically, I await a signal for a certain amount of time until either a connection is made to the server, or the awaiter times out. If connection has been made, then great, we continue with the tests. If not, then we abort the tests.

Seems to work fine. I'll stick with this for the time being unless there's a better solution.

LowFire:
> I think this is a good addition. This should probably also be done for the post-run script. And maybe `should_skip_script` per the idea I have below.
> 
[...]

PR #671, opened 2024-11-28
add await in pre-run script, post-run scrpt, and should_skip_script.

LowFire:
In response to #670


bitwes:
Thanks for the PR, well done on the tests.

Issue #673, opened 2024-12-02
Font import files are changed by Godot editor 4.3


manuq:
# Versions
* Godot: 4.3
* GUT: 9.3.0
* OS: Endless OS

# The Bug
The import files for fonts `addons/gut/fonts/*.ttf.import` are changed by Godot 4.3 because the engine adds a `disable_embedded_bitmaps=true` in them.

# Steps To Reproduce
In a versioned project, install Gut. Then commit the files as released. Then use Gut. You will see the following files being modified:
```
M	addons/gut/fonts/AnonymousPro-Bold.ttf.import
M	addons/gut/fonts/AnonymousPro-BoldItalic.ttf.import
M	addons/gut/fonts/AnonymousPro-Italic.ttf.import
M	addons/gut/fonts/AnonymousPro-Regular.ttf.import
M	addons/gut/fonts/CourierPrime-Bold.ttf.import
M	addons/gut/fonts/CourierPrime-BoldItalic.ttf.import
M	addons/gut/fonts/CourierPrime-Italic.ttf.import
M	addons/gut/fonts/CourierPrime-Regular.ttf.import
M	addons/gut/fonts/LobsterTwo-Bold.ttf.import
M	addons/gut/fonts/LobsterTwo-BoldItalic.ttf.import
M	addons/gut/fonts/LobsterTwo-Italic.ttf.import
M	addons/gut/fonts/LobsterTwo-Regular.ttf.import
```

PR #674, opened 2024-12-02
Font import files: Add disable_embedded_bitmaps

manuq:
The Godot editor 4.3 adds this by default.

https://github.com/bitwes/Gut/issues/673


bitwes (approved):
Thanks, I noticed this on an import but hadn't paid any attention to why the files changed.

PR #687, opened 2025-01-16
Adding support for dictionaries in parameters

### LINKED ISSUE #686: Improve parametrized test UX
# Versions

4.4-dev7

# The Feature

I'm currently generating 106 test asserts for 20 parameter entries and the UX just lists all the failed asserts without any sort of grouping:

Image

It would be nice if there was a way to group these entries, ideally via a text label so that it is easier to debug what's actually failing.

ghost:
*Fixes #686:* This PR adds support for using dictionaries in parameters instead of arrays for a better UX. The user will be able to pass a dictionary to the `use_parameters()` function and receive the keys instead of the data.

Please let me know if this path is desirable and I'll add relevant tests and user documentation.

Example:

```gdscript
var _height_unit := sqrt(0.75)
var _hex_id_params := {
	"0:0": [Vector2(0,0), Vector2i(0,0)],
	"0:1": [Vector2(0,2*_height_unit), Vector2i(0,1)],
	"0:-1": [Vector2(0,-2*_height_unit), Vector2i(0,-1)],
	"0:2": [Vector2(0,4*_height_unit), Vector2i(0,2)],
}

func test_coordinates_to_hex_id(param_name=use_parameters(_hex_id_params)) -> void:
	var params : Array = _hex_id_params[param_name]
	#...
```

Additionally, the result will print the keys in the result instead of the indexes:

image

### DIFF
--- a/addons/gut/collected_test.gd
+++ b/addons/gut/collected_test.gd
@@ -114,5 +114,3 @@ func to_s():
 	for i in range(pass_texts.size()):
 		to_return += str(pad, 'Pass:  ', pass_texts[i], "\n")
 	return to_return
-
-
--- a/addons/gut/gut.gd
+++ b/addons/gut/gut.gd
@@ -846,6 +846,9 @@ func get_call_count_text():
 		# should have been called before we get to any calls for this method
 		# just due to how use_parameters works.  There isn't a way to know
 		# whether we are before or after that call.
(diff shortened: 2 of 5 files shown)


bitwes:
You can use the `ParameterFactory` to create "named parameters".  Check out https://gut.readthedocs.io/en/latest/Parameterized-Tests.html#named-parameters

If that doesn't address the issue, let me know how this is different.

PR #692, opened 2025-02-20
Fix docs typo

Curtis-Barnhart:
Just a few typos here and there, and one clarification of a sentence I thought was a little confusing to read.

I also added a small bit to the `generate_rst.sh` file so that I could build it on my linux machine (otherwise it would use the mac version of a command that I don't have here).

### DIFF
--- a/documentation/generate_rst.sh
+++ b/documentation/generate_rst.sh
@@ -44,11 +44,11 @@ function generate_xml(){
     # using gtimeout (which is mac version of timeout from coreutils) and then
     # kill it (-k 1s).
     case $OSTYPE in
-        linux-gnu)
-            timeout -k 1s 2s $GODOT --doctool $the_dir --no-docbase --gdscript-docs $scripts_dir
-        ;;
-        darwin)
+        "darwin"*)
             gtimeout -k 1s 2s $GODOT --doctool $the_dir --no-docbase --gdscript-docs $scripts_dir
+        ;;
+        *)
+            timeout -k 1s 2s $GODOT --doctool $the_dir --no-docbase --gdscript-docs $scripts_dir
         ;;
     esac
 
--- a/documentation/generate_rst.sh
+++ b/documentation/generate_rst.sh
@@ -43,7 +43,14 @@ function generate_xml(){
     # soon (fixed merged after 4.3).  So we wait 2 seconds +1 seconds (-k 1s)
     # using gtimeout (which is mac version of timeout from coreutils) and then
     # kill it (-k 1s).
-    gtimeout -k 1s 2s $GODOT --doctool $the_dir --no-docbase --gdscript-docs $scripts_dir
+    case $OSTYPE in
+        linux-gnu)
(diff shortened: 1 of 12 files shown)


documentation/generate_rst.sh
```diff
@@ -43,7 +43,14 @@ function generate_xml(){
     # soon (fixed merged after 4.3).  So we wait 2 seconds +1 seconds (-k 1s)
     # using gtimeout (which is mac version of timeout from coreutils) and then
     # kill it (-k 1s).
-    gtimeout -k 1s 2s $GODOT --doctool $the_dir --no-docbase --gdscript-docs $scripts_dir
+    case $OSTYPE in
```
bitwes, line 46: Your original case statement didn't work for me because `$OSTYPE` has a version in it (at least for darwn); my `$OSTYPE` is "darwin23.0".  This code changes the OS check to `"darwin"*` and then uses `timeout` as the default for any other OS.  As I was testing this I figured `timeout` will probably work on most other shells and therefore should be the default case.  

Please test this locally as I don't have a linux box to test on and could have broken it again.  I've been testing by doing the following:
``` zsh
# kill all rst files
rm documentation/docs/class_ref/*.rst
# look to see that they are all deleted because I never trust what I do
git status
# regenerate them
zsh documentation/generate_rst.sh
# make sure they are all back
git status
```

Here's the change to the comment and code:
```zsh
    # The --doctool command always hangs indefinitely. We use timeout to kill it
    # after some time.  The following case statement matches the timeout method
    # to the OS (basically Mac is special).  If your shell has a differently 
    # named timeout then add it to the case statement.
    #
    # The "hangs forever" bug might be fixed soon (a fix was merged after 4.3).
    case $OSTYPE in
        "darwin"*)
            gtimeout -k 1s 2s $GODOT --doctool $the_dir --no-docbase --gdscript-docs $scripts_dir
        ;;
        *)
            timeout -k 1s 2s $GODOT --doctool $the_dir --no-docbase --gdscript-docs $scripts_dir
        ;;
    esac
```

bitwes (changes requested):
Thanks for the PR.  It looks like you got the documentation generation working (original PR generated email indicated you were getting an error).  I really appreciate that.  I wasn't sure if any of the documentation I made would make any sense to anyone else.  I get confused every time I get back into it.  I'd like to hear any feedback about contributing to the documentation.

I have one change to `generate_rst.sh` that I'd like you to make and test.  Everything else looked good.

Curtis-Barnhart:
Yeah, it turned out that SELinux was getting in the way of the docker container properly accessing the mounted docs directory. There were some strange file properties I had to change somewhere and then it all worked.
Reading the documentation/contributing to it was pretty straightforward. All the instructions worked out of the box for me (aside from having to figure out that it expected the environment variable GODOT to be set, and aside from the SELinux thing, which was a problem on my end). I'll probably be looking into the docs more deeply in the near future as I actually start to mess around with using Gut more.

documentation/generate_rst.sh
Curtis-Barnhart: The change works perfect over here.

Issue #576, opened 2024-03-02
c# support


bitwes:
Issues that need to be resolved to test c# from gdscript
- [ ] `inst_to_dict` does not work on c# scripts.
- [ ] GDScript cannot extend a C# script and vice versa.  This complicates doubling/stubbing/spies.

Issues that need to be resolved to test c# from c#
- [ ] Cannot extend `GutTest` in C#, so there will have to either be a wrapper for all of `test.gd` or a C# class with a `test.gd` instance that has been setup.
- [ ] The C# people will want to name their tests in PascalCase.  This might be ok, since test prefix is configurable.  But if you want to write tests in both C# and GDScript then we'll need a prefix for both.  Also, I hate that C# likes PascalCase instead of camelCase for function names.  Really hate it.

ganondev:
In response to the question on the PR, my primary motivation for pursuing C# in GUT rather than just writing tests in something like NUnit is that I seek the sophisticated engine integrations build into GUT's constructs. Gut runs inside of a Godot process, it's not standalone. In contrast, C# testing libraries could really only orchestrate around Godot, and maybe some utilties could be written to interact with the engine for context, but it would be complex to inject text fixtures into the engine from an external process and tie the engine lifecycle to the test framework's. Without significant work, using C# frameworks one can only hope to test C# code directly in an engine-agnostic kind of way. One would not get the value of more thorough "integration" tests of C# resources in the engine. This can work, as I have done it before for a separate godot project, but I learned there are limitations on what kind of guarantees you can set yourself up for.

namelessvoid:
The `function` recorded in `GutTrackedError` behaves differently when pushing errors via `GD.PushError()` from C#. In GDScript, the recorded function is `push_error`. In C#, the `function` is set to the signature of the C# method that calls `GD.PushError()`.

Example:

```c#
namespace WildLifeSpot.AnimalSpots.Domain;

[GlobalClass]
public partial class AnimalSpotReportQueryHandler: Node
{
    public Variant GetAnimalSpotReport(AnimalSpotReportQuery query)
    {
        if (!query.IsValid())
        {
            GD.PushError("Received invalid query. Returning empty report.");
            return Variant.From(AnimalSpotReport.Empty());
        }
   }
}
```
will set `function` to 
```
[...]

PR #700, opened 2025-04-13
poc run C# tests

ganondev:
Quick and dirty proof of running tests written in C#. The approach to solving the problem of "inheriting" from the base GutTest class is to create a C# version which is in theory indestinguishable from the GdScript version as an interface but in reality proxies any relevant lifecycle method calls, except those that are meant to be overridden, to an inner private instance of the GdScript `GutTest` node. There is also some additional utlitity around collecting and running the C# tests.

Currently this has been tested by running from the cli. This is still quite early, certainly a lot of edges to cover, but at least it "works" insofar as the C# tests are executed and passing (with warnings about orphans and such).


bitwes:
Very cool.  It's great to see some of the work done to get an idea of what it will take.  You got a lot of the basics proven out.  I think the next thing to try out is creating doubles in C#.  If that can be figured out then we might be on to something.

Can you comment on #576 with any reasons you have for wanting to use GUT and C# instead of using some other C# testing framework?

ganondev:
Regarding doubles, I've explored a few approaches, and based on what I can tell:
- Doubling as it currently functions (applied to gdscript, native, and scenes) will continue to work normally. Leveraging this in C# tests works out of the box, it just means you have to manually `Call` doubler methods which might be tedious without a wrapper.
- Regardless of the language used for tests themselves, doubling C# scripts in particular would be complicated. This is because doubling scripts, whether native or gdscript, requires runtime-generated subclasses of the target. It's not possible to extend C# in GDScript, so the options would be:
  - For C# specifically, generate wrappers instead of subclasses. I'm not sure of the functionality impacts of this approach. This could in theory just be a new strategy in the `DOUBLE_STRATEGY` enum, which is forced for C# scripts like `INCLUDE_NATIVE` is for native nodes.
  - For C# scripts, generate and dynamically proxy method calls using something like `DynamicProxy` (https://github.com/castleproject/Core/blob/master/docs/dynamicproxy.md). Unlike the GDScript generation, this would happen in-memory, managed by the mono clr. For what it's worth, this is apparently the approach that all C# mocking libraries use. Taking this a step further, a C# double could just wrap one of these libraries.

Either way, it's a decent consideration for the direction you'd want to take your codebase in. Despite being the "standard" approach for C# in other places, the latter means a completely separate doubler implementation for C#, and likely some licensing considerations, whereas the former means a big refactor to how doubling works, with some fundamental limits on functionality (off the top of my head, inheritance checks wouldn't work).

Thoughts?

ganondev:
I've come to find out that passing a C# object to some methods like asserts results in the `inst_to_dict` failure pointed out in the original issue, so I've added it to the list. At the moment, you can't pass C# objects as args to (some/all?) asserts. Not exhaustive but just something I've run into.

PR #702, opened 2025-04-17
Wrap skipped and failure messages in CDATA

ali-wallick:
Ran into an issue with parsing the XML of the exported test results when the error message has null in it as in Godot `str(null)` is `<null>`:

```
<failure message="failed">params[10] Expected [<null>] to be anything but NULL:  [message]
            at line -1</failure>
```

I was able to solve the issue locally with this change, but not sure if there's another route you'd prefer (eg escaping certain characters).

### DIFF
--- a/addons/gut/junit_xml_export.gd
+++ b/addons/gut/junit_xml_export.gd
@@ -10,6 +10,10 @@ func indent(s, ind):
 	to_return = to_return.replace("\n", "\n" + ind)
 	return to_return
 
+# Wraps content in CDATA section because it may contain special characters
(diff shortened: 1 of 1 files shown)


bitwes:
This looks good.  Thank you.  

I wanted to add some tests to replicate the issue and validate the fix.  I tested your code with the tests and everything passed.  

In order to combine my tests and your code I changed the branch this PR is attempting to merge into.  I've never done that before, but it seemed like a faster way to get all the code in one place and to make sure you are credited.

ali-wallick:
That's great, thank you!

Issue #667, opened 2024-11-04
--script (rather than -s) no longer supported


wjt:
# Versions

* Godot: 4.3.stable
* GUT: 9.3.0
* OS: Linux

# The Bug

The new option parser in GUT 9.3.0 only understands the `-s` short form argument to `godot`, not the `--script` long form.

# Steps To Reproduce

In a project with GUT installed, run:

```
godot --script addons/gut/gut_cmdln.gd
```

The expected result is that the tests run. The actual result is that the following is printed to the terminal:

```
Godot Engine v4.3.stable.flathub.77dcf97d8 - https://godotengine.org
Inconsistent value (1) for DRI_PRIME. Should be < 1 (GPU devices count). Using: 0
OpenGL API 4.6 (Core Profile) Mesa 24.2.5 (git-3b9fcb7e4d) - Compatibility - Using Device: Intel - Mesa Intel(R) UHD Graphics 620 (WHL GT2)

Unknown arguments:  ["--script", "addons/gut/gut_cmdln.gd"]
```

Changing the command line to the following works around the problem:

```
godot -s addons/gut/gut_cmdln.gd
```

In older versions, the function `get_unused_options()` had the following logic:

```gdscript
          var script_option = to_return.find("-s")
          if script_option == -1:
                  script_option = to_return.find("--script")
          if script_option != -1:
                  to_return.remove_at(script_option + 1)
                  to_return.remove_at(script_option)
```

In 9.3.0, the `Options` class has:

```gdscript
	var script_option = Option.new('-s', '?', 'script option provided by Godot')

# ...
	func get_by_name(option_name):
		var found_param = null
		if(option_name == script_option.option_name):
			found_param = script_option
# ...
```

which does not handle the `--script` name.

bitwes:
This should be fixed now.

PR #705, opened 2025-04-18
Optparse option aliases

Curtis-Barnhart:
#667 describes a regression where `addons/gut/gut_cmdln.gd` stopped recognizing the `--script` argument. This PR fixes this problem and also introduces argument aliases to the optparse system.
Any option can now be added with any number of aliases. Aliases will be checked to make sure they do not collide with other option names or aliases in the same way that option names are usually checked for collisions. An option's aliases will be printed out at the end of its help string.

### DIFF
--- a/addons/gut/cli/optparse.gd
+++ b/addons/gut/cli/optparse.gd
@@ -196,7 +196,7 @@ class Options:
 	var default_heading = OptionHeading.new()
 	var script_option = Option.new('-s', '?', 'script option provided by Godot')
 
-	var _options_by_name = {}
+	var _options_by_name = {"--script": script_option}
 	var _options_by_heading = [default_heading]
 	var _cur_heading = default_heading
 
--- a/addons/gut/cli/optparse.gd
+++ b/addons/gut/cli/optparse.gd
@@ -219,6 +219,7 @@ class Options:
 		if aliases != null:
 			for a in aliases:
 				_options_by_name[a] = option
+			option.aliases.assign(aliases)
 
 
 	func add_positional(option):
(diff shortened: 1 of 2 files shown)


bitwes:
I've been wanting this feature for a bit.  This will be great to have.

`find_custom` is a 4.4 feature and GUT requires 4.3 (also I'm using 4.3 still).  I don't think this is a big enough feature to require 4.4 but I understand not wanting to work in 4.3.  So, please either:
*  Change the implementation to work in Godot 4.3 
or 
* Make the PR against the [Godot 4.4 branch](https://github.com/bitwes/Gut/tree/godot_4_4).

Curtis-Barnhart:
Oh dear silly me. Probably should've thought about that first. Here's a fixed version for 4.3.

bitwes:
Sorry for the delay.  I checked this out a bit then got distracted with Godotcon.  This works, but I think I would like the following changes:
* Change the handling of `op_name` in the `add` function to handle a string value or an array.
* When it is a string, you get a an option without an aliases.
* When it is an array, then every value after the first is an alias.

This allows us to keep all the option names together which is easier to read.  In my experience it is rare that you would want more than one alias for an option (I can't think of an example off the top of my head).  By putting the optional array at the front, we can still support more aliases but don't have to use a single element array in usual use case.

This would change the usage to something like:
```gdscript
op.add("--name", "default", "description")
op.add(["--name", "-n"], "default", "description")
op.add(["--name", "--nombre", "--namna", "-n"], "default", "description")
```

Curtis-Barnhart:
Okay, it's been updated to have the specified behavior. I also added type hints and docstrings to all the user facing methods and rebuilt the class reference docs for them.

bitwes:
This looks good.  Thanks a lot for adding this.

PR #711, opened 2025-05-02
Docstrings and type hints for junit_xml_export and strutils

Curtis-Barnhart:
# Docstrings & Type Hints

Adds docstrings and type hints to all methods in addons/gut/junit_xml_export.gd and some in addons/gut/strutils.gd. Does not add type hints wherever a union was expected since Godot does not support this. I'm not sure if you _want_ type hinting in this project, since I don't really see it much elsewhere in the codebase, so I can definitely remove it if you'd prefer. I just find it to be quite helpful as its own form of documentation and for code completion.

# Question: Static Members

Nothing in this class records any sort of state, and I'm a little confused as to why its methods are not static. Because they are not static, this class must be instantiated every time someone wants to call one of its methods, which seems unnecessary. If there's not a reason you have specifically avoided doing this, I'd also like to make this class' methods static in this PR (which would involve updating its usage everywhere it is invoked, of course).


bitwes:
# Short answer:
* Type hints are fine for user facing features, but maybe not internally due to existing precedence.
* Don't add docstrings to private methods/variables.
* `strutil` changes are good.
* Making `strutils` static is probably fine but I'd rather not.

# Long Answers:
## Type Hints
GUT has been around a long time and type hints became more popular after I was already very happy being lazy.  When I'm in a dynamically typed language, I like to be lazy.  That being said, I have been using type hints more in personal projects just for the autocomplete.  Adding it to user facing methods is ok with me.  Also adding it to an entire file (especially small ones) is fine as well, so adding them to `junit_xml_export` is good.  (I probably need a style guide where I could state this).

## Docstrings
We shouldn't add docstring comments to private methods/variables because they will show up in the editor's documentation and we don't want that.  The `@ignore` annotation is a custom one I made for the wiki generated documentation.  Godot doesn't have this feature so they will show up there.  In addition to this, I try not to comment private methods unless it is absolutely necessary.  Private functions/variables should be easy to change and comments make more work.

## Static Classes in GUT
Making `strutils` static wasn't done because static classes aren't always available early in running a project.  GUT didn't have any static classes, autoloads, or use `class_name` due to when they are loaded versus when they are referenced when running GUT (mostly from the command line/vscode).  I think I've finally worked around those kinds of issues but I'm always wary of adding more static classes or `class_name`.  I could probably be convinced to change it but I would bring that up in a different pull request.

Curtis-Barnhart:
I've updated the docstrings according to your directions. Does this look good?

bitwes:
Sorry, I thought I was waiting on you.  I'm ok with these changes.  Thank you!

PR #713, opened 2025-05-02
Clear embedded font from OutputText.tscn

gbrock:
I noticed an unusually large scene in the `gui/` directory of the addon: OutputText.tscn, at 8.75 MB.  I found an embedded font file `Courier Prime` and some image data that bloated that scene.  By setting the font of the `Output` node to the font file already included in the project, **that file is now only 9 KB**.  So now the whole addon goes from ~11MB to ~2MB to install.


bitwes:
Great find, thanks!

Issue #709, opened 2025-04-30
GUT broken in Godot 4.5 dev3


mosquitobird11:
# Versions
*(list all versions where you have replicated the bug)*
* [Godot: 4.5 Dev 3](https://godotengine.org/article/dev-snapshot-godot-4-5-dev-3/)
* GUT: 9.3.1
* OS: Windows 11 Home

# The Bug
 - I get many errors on load, like:
```bash
  ERROR: res://addons/gut/utils.gd:101 - Parse Error: The member "Logger" shadows a native class.
  ERROR: res://addons/gut/gui/OutputText.gd:0 - Compile Error: Failed to compile depended scripts.
  ERROR: modules/gdscript/gdscript.cpp:3023 - Failed to load script "res://addons/gut/gui/OutputText.gd" with error "Compilation failed".
  ERROR: res://addons/gut/utils.gd:101 - Parse Error: The member "Logger" shadows a native class.
  ERROR: res://addons/gut/utils.gd:101 - Parse Error: The member "Logger" shadows a native class.
  ERROR: res://addons/gut/gut_to_move.gd:0 - Compile Error: Failed to compile depended scripts.
  ERROR: res://addons/gut/utils.gd:101 - Parse Error: The member "Logger" shadows a native class.
  ERROR: res://addons/gut/gut.gd:0 - Compile Error: Failed to compile depended scripts.
  ERROR: res://addons/gut/test.gd:0 - Compile Error: Failed to compile depended scripts.
  ERROR: modules/gdscript/gdscript.cpp:3023 - Failed to load script "res://addons/gut/test.gd" with error "Parse error".
  ERROR: res://addons/gut/utils.gd:101 - Parse Error: The member "Logger" shadows a native class.
  ERROR: res://addons/gut/gui/gut_config_gui.gd:0 - Compile Error: Failed to compile depended scripts.
  ERROR: modules/gdscript/gdscript.cpp:3023 - Failed to load script "res://addons/gut/gui/gut_config_gui.gd" with error "Parse error".
  ERROR: res://addons/gut/utils.gd:101 - Parse Error: The member "Logger" shadows a native class.
  ERROR: res://addons/gut/gut_config.gd:0 - Compile Error: Failed to compile depended scripts.
```

I believe this is due to this PR now merged to godot `master` branch: https://github.com/godotengine/godot/pull/91006 
> Additionally, added a custom Logger class to allow users intercept the internal messages

# Steps To Reproduce
 - Load the addon
 - Try to do anything

PR #715, opened 2025-05-19
Avoid global `Logger` as this conflicts with a native class in Godot 4.5

dreed-sd:
Fixes https://github.com/bitwes/Gut/issues/709

```
ERROR: res://addons/gut/utils.gd:95 - Parse Error: The member "Logger" shadows a native class.
```


dreed-sd:
The `Logger` commit fyi: https://github.com/godotengine/godot/pull/91006

murilosr:
Worked great here! Thanks

bitwes:
Thanks!  I was a little hesitant to merge it since it seem sodd they would take the name `Logger` from everyone.  The more I thought about it though, it would be good to not use that name in Gut regardless of whether Godot keeps using it.

bitwes (approved):

PR #716, opened 2025-05-20
Fix Errors during Godot 4.4 Editor Startup

garettbass:
This change fixes three errors I was encountering during Godot 4.4 Editor startup.  The errors didn't seem to prevent Gut from working correctly.

```
SCRIPT ERROR: Invalid call. Nonexistent function 'new' in base 'GDScript'.
at: @static_initializer (res://addons/gut/utils.gd:160)

SCRIPT ERROR: Invalid call. Nonexistent function 'error_if_not_all_classes_imported' in base 'GDScript'.
at: _init (res://addons/gut/gut_plugin.gd:7)

SCRIPT ERROR: Invalid call. Nonexistent function 'error_if_not_all_classes_imported' in base 'GDScript'.
at: _version_conversion (res://addons/gut/gut_plugin.gd:15)
```

Gut does a great job of unit testing my code.  Thanks for all of your hard work on this project.


dagerydoo:
I just hit an empty settings panel trying out GUT for the first (literally just installed GUT to play with this evening). This change fixed it, FWIW. Thanks for the fix, and I look forward to unit testing with GUT, thanks for the project!

murilosr:
I applied the merge here, and it worked with my project, too. Thanks!

bitwes:
Pleas open up an issue where the symptoms of this can be tracked.  I'd like to have the issue better described somewhere else.  Please include info about:
- OS and Godot version
- Do the errors reported show up at every launch or just on new install?
- Where do the errors appear?

The `class_name`s should be prefixed with "Gut" for name-spacing concerns.

Can someone confirm that this fixes the issue on a fresh install of GUT?  This will require deleting the `.godot` folder.   There are some initial errors when opening a project for the first time after installing GUT.  I'm not able to replicate the issue reported here with a new 4.4.1 project.

Thanks.  These startup errors and warnings are a pain to deal with, all the info you can provide might help prevent future ones.

murilosr:
@bitwes I remember that these errors started when I updated from 4.3 to 4.4 (I don't remember which version, I think it was some of 4.4-dev). My project had other initialization errors (from NavigationServer and some _enter_trees from some scripts), so I just started ignoring them all.
After this merge, I know that the GUT-related errors stopped.

jammons:
I opened a related issue here: https://github.com/bitwes/Gut/issues/724. (sorry for the messiness, I ran into two issues while trying to set up GUT -- I can close/reopen if needed).

I tried the patch suggested here and unfortunately am still seeing the same error messages with the changes in place. I tried both reloading my project and toggling GUT on and off in the project plugin config.

bitwes:
I have not been able to replicate this issue locally.  I'm not sure why the PR fixes the issue and could introduce different bugs for the command line.  If anyone can provide a project and steps to reproduce that would be very helpful.  Also include your OS when you post it.

bitwes:
#723 might have a good lead on this issue.  Also, if you could install GUT 9.4.0 which was recently released and report back any findings that would be very helpful.

m-radzikowski:
I've created #731 with the issue. Unfortunatelly, the patch from this PR does not fix it for me.

From my tests it seems to happen only if LSP connection from VSCode is made before GUT initialization. Which is weird but would explain race condition. Can someone else confirm if they are using VSCode too? (maybe best in issue comments, not here)

garettbass:
I’m using VSCode, so this seems likely.On Jul 9, 2025, at 11:08 AM, Maciej Radzikowski ***@***.***> wrote:﻿m-radzikowski left a comment (bitwes/Gut#716)
I've created #731 with the issue. Unfortunatelly, the patch from this PR does not fix it for me.
From my tests it seems to happen only if LSP connection from VSCode is made before GUT initialization. Which is weird but would explain race condition. Can someone else confirm if they are using VSCode too? (maybe best in issue comments, not here)

bitwes:
I'm going to close this PR. I have deemed #731 as the official issue for this issue.  Adding `class_name` in various places is not a good solution to the issue.  This appears to be a Godot/LSP issue that has been fixed in 4.5.

P.S. You can run tests through VSCode and completely ignore this issue when it happens.

P.S.S. @garettbass your icon looks like Ryu from Ninja Gaiden whenever I see it out of the corner of my eye.  I've been consistently confused for a second or two when I look directly at your icon.  It takes way too long for me to realize I'm looking at glasses and not the face of a ninja with a blue balaclava on.  Good job.

P.P.P.S. I don't think you should add additional "S"s to additional postscripts.  I think you should be adding additional "P"s.

PR #737, opened 2025-08-05
Godot 4 5 error hook

bitwes:
See changes to CHANGES.md, it's a bunch of error related stuff.


Toxe:
Nice, I quite like the new error handling options! I am testing a database class and in 4.4 it would just write the error into the log but now I can actually quite easily assert the expected error and then ignore it.

Well done!

```gdscript
extends GutTest

var db: Database = null

# [...]

func assert_and_ignore_sqlite_error(expected_sqlite_error: String) -> void:
    var errors: Array = get_errors()
    assert_gt(errors.size(), 0)

    for err: GutTrackedError in get_errors():
        var err_msg: String = err.code
        if err_msg.contains(expected_sqlite_error):
            err.handled = true
            return

    fail_test("expected SQLite to raise error \"%s\"" % expected_sqlite_error)


func test_select_rows_returns_false_on_database_error() -> void:
    assert_false(db.select_rows("missing_table", "id=99", ["col1", "col2"]))
    assert_and_ignore_sqlite_error("no such table: missing_table")
```

bitwes:
@Toxe Thanks!  This is neat to see in the wild already.  Does `select_rows` generate more than one error with the expected text?  If not you should just be able to do 
```gdscript
assert_engine_error("no such table: missing_table)
# Or assert_push_error if the are generated from push_error(...)
```
instead of looping over all the errors.  The asserts prevent failure due to the expected error and will also fail if the error cannot be found.

Toxe:
@bitwes Ooooh! I did not see that! Also I just noticed that there is a new "Error Tracking" page in the docs. Great! This does indeed do everything I need.

Guess I should have continued reading a second longer after finding `get_errors()`. 🤦 😆

PR #739, opened 2025-08-15
Godot 4 5 better orphans

bitwes:
* Orphaned nodes are now listed instead of just counted.
* Children of orphaned nodes are counted to cut down on clutter.
* All orphans are listed at the end of the run.
* Fixed a few reference leaks.
* Creating doubles of RefCounted objects no longer causes a leaked reference.
* No orphans are generated in GUT tests.


addons/gut/script_parser.gd
bitwes, line 1: Why did I do this and can I not do this?

test/unit/test_bugs/test_i608_mouse_motion_buttons.gd
```diff
@@ -1,5 +1,10 @@
 extends GutTest
 
+func should_skip_script():
```
bitwes, line 3: Put this in GutInternalTester as a method?

addons/gut/script_parser.gd
bitwes: I think this is because `__gutdbl` is now cleared at the end of a test but if either of these two methods are included then when they are called as part of normal cleanup, `__gutdbl` is null.

PR #750, opened 2025-08-28
fix: register_inner_classes can handle cyclic dependencies

thephilipquan:
If you have the following:
```gdscript
# script A
const B := preload("res://b.gd")

# script B
const A := preload("res://a.gd")
```
Calling `register_inner_classes` on `A` results in a stack overflow.

### DIFF
--- a/addons/gut/inner_class_registry.gd
+++ b/addons/gut/inner_class_registry.gd
@@ -19,13 +19,7 @@ func _register_inners(base_path, obj, prev_inner = ''):
 		var key = consts[const_idx]
 		var thing = const_map[key]
 
-		if(typeof(thing) == TYPE_OBJECT):
-			if (
-				thing.resource_path == base_path
-				or thing in _registry
-			):
-				const_idx += 1
-				continue
+		if(typeof(thing) == TYPE_OBJECT and thing.resource_path == ''):
 			var cur_inner = str(prev_inner, ".", key)
 			_registry[thing] = _create_reg_entry(base_path, cur_inner)
 			_register_inners(base_path, thing, cur_inner)
--- a/addons/gut/inner_class_registry.gd
+++ b/addons/gut/inner_class_registry.gd
@@ -20,6 +20,12 @@ func _register_inners(base_path, obj, prev_inner = ''):
 		var thing = const_map[key]
 
 		if(typeof(thing) == TYPE_OBJECT):
+			if (
+				thing.resource_path == base_path
+				or thing in _registry
+			):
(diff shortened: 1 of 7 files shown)


addons/gut/inner_class_registry.gd
```diff
@@ -20,6 +20,12 @@ func _register_inners(base_path, obj, prev_inner = ''):
 		var thing = const_map[key]
 
 		if(typeof(thing) == TYPE_OBJECT):
```
bitwes, line 22: Change this line to:
```
if(typeof(thing) == TYPE_OBJECT and thing.resource_path == ''):
``` 
Remove all other changes.

addons/gut/script_parser.gd
bitwes, line 1: Remove the changes to this file.

bitwes (changes requested):
Nice find.  I looked into this a bit, and I think the actual issue is that I'm not ignoring references to external scripts.  I think we only need to change one line so that it doesn't include `const` script references.  I tested this out a little and it seemed to work.  The change is marked in the review.

I used the following tests, added to `test_inner_class_registry.gd` to verify the fix:
```gdscript
var CyclicRefA = load('res://test/resources/parsing_and_loading_samples/cyclic_ref_class_a.gd')

func test_does_not_spin_out_of_control_with_const_cyclic_refs():
	var reg = GutUtils.InnerClassRegistry.new()
	reg.register(CyclicRefA)
	pass_test('we got here')

func test_does_not_contain_a_ref_to_external_classes():
	var reg = GutUtils.InnerClassRegistry.new()
	reg.register(CyclicRefA)
	var result = reg.get_base_path(CyclicRefA.b_ref)
	assert_null(result)
```
I also added two scripts to replicate the issue, used in the tests.

cyclic_ref_class_a.gd
```gdscript
const b_ref := preload('res://test/resources/parsing_and_loading_samples/cyclic_ref_class_b.gd')

class CyclicRefAInnerClass:
	var foo = 'bar'

	class CyclicRfAInnerInnerClass:
		var bar = 'foo'
```

cyclic_ref_class_b.gd
```gdscript
const a_ref := preload('res://test/resources/parsing_and_loading_samples/cyclic_ref_class_a.gd')

class CyclicRefBInnerClass:
	var foo = 'bar'
```

Please add these into your PR and any other tests you can think of.  I didn't think too hard on the problem, just enough to replicate and review.

Thanks!

thephilipquan:
Thanks for taking a look at this. I added your changes and tests.

Issue #746, opened 2025-08-26
Godot 4.5 input tests fail in non-floating in-editor run.


bitwes:
# Versions
*(list all versions where you have replicated the bug)*
* Godot: 4.5 beta6
* GUT: godot_4_5
* OS: Mac

# The Bug
GUT test:  res://test/unit/test_illustrate_input_and_button_event_issue.gd
Some tests fail when run through the editor and the "Game Workspace" is not a floating window.  

All tests pass when "Game Workspace" is floating or through either of the new "External" options.

```
res://test/unit/test_illustrate_input_and_button_event_issue.gd
* test_illustrate
    [Failed]:  Expected [_on_button_up] to have been called on @Button@6:<Button#78802585848>(partial-double of test_illustrate_input_and_button_event_issue.gd/PrintEventsButton)
    Calls made on @Button@6:<Button#78802585848>
      * _ready()
      * _input(InputEventMouseButton: button_index=1, mods=none, pressed=true, canceled=false, position=((42.69302, 42.01318)), button_mask=0, double_click=false)
      * _input(InputEventMouseButton: button_index=1, mods=none, pressed=false, canceled=false, position=((42.69302, 42.01318)), button_mask=0, double_click=false)
      * _unhandled_input(InputEventMouseButton: button_index=1, mods=none, pressed=true, canceled=false, position=((42.69302, 42.01318)), button_mask=0, double_click=false)
      * _unhandled_input(InputEventMouseButton: button_index=1, mods=none, pressed=false, canceled=false, position=((42.69302, 42.01318)), button_mask=0, double_click=false)
      at line 86
    [Failed]:  Expected object @Button@6:<Button#78802585848>(partial-double of test_illustrate_input_and_button_event_issue.gd/PrintEventsButton) to have emitted signal [button_down]:   (Signals emitted: ["minimum_size_changed", "draw"])
      at line 87
    [Failed]:  Expected object @Button@6:<Button#78802585848>(partial-double of test_illustrate_input_and_button_event_issue.gd/PrintEventsButton) to have emitted signal [button_up]:   (Signals emitted: ["minimum_size_changed", "draw"])
      at line 88
* test_same_thing_the_gut_way
    [Failed]:  Expected object @Button@7:<Button#81621158155>(test_illustrate_input_and_button_event_issue.gd/PrintEventsButton) to have emitted signal [button_down]:   (Signals emitted: ["minimum_size_changed", "draw"])
      at line 103
[...]

PR #763, opened 2025-09-21
Add abstract class/method support

### LINKED ISSUE #761: Doubling an abstract class with abstract method fails with error
# Versions

* Godot: v4.5.stable.mono.official [876b29033]
* GUT: 9.5.0
* OS: Ubuntu 25.10

# The Bug

I have an @abstract class with @abstract methods:

```gdscript
@abstract
class_name FSCameraRepository
extends Node

@abstract
func find_all() -> Array[FSCamera]

@abstract
func save(p_camera: FSCamera) -> void

@abstract
func delete(p_id: int) -> void
```

When I try to `double` this in a test like this:

```gdscript
extends GutTest

func test_double_abstract_class():
	var repo_double = double(FSCameraRepository).new()
```

then I see this error:

`Parser Error: Cannot call the parent class' abstract function "find_all()" because it hasn't been defined.`

# Steps To Reproduce

Copy the class and test from above and try to execute the test.

namelessvoid:
Resolves #761 

This PR implements abstract class/method support. It does this with two changes:

1. The `MethodMaker` and `function_template.txt` is changed to not create a call to `super()` if the parent method is abstract. Since this code path should not be called (see 2.) it was replaced by a `assert(false)` to indicate a bug in GUT.
2. The `StubParams.to_call_super()` was updated, too, to include checking the parent's method. If it is abstract, it logs an error and returns.

### DIFF
--- a/addons/gut/doubler.gd
+++ b/addons/gut/doubler.gd
@@ -207,8 +207,8 @@ func _get_inst_id_ref_str(inst):
 	return ref_str
 
 
-func _get_func_text(method_hash):
-	return _method_maker.get_function_text(method_hash) + "\n"
+func _get_func_text(method_meta):
+	return _method_maker.get_function_text(method_meta) + "\n"
 
 
 func _parse_script(obj):
--- a/addons/gut/doubler.gd
+++ b/addons/gut/doubler.gd
@@ -207,8 +207,8 @@ func _get_inst_id_ref_str(inst):
 	return ref_str
 
 
-func _get_func_text(method_meta):
-	return _method_maker.get_function_text(method_meta) + "\n"
+func _get_func_text(method_hash):
+	return _method_maker.get_function_text(method_hash) + "\n"
 
 
 func _parse_script(obj):
--- a/addons/gut/script_parser.gd
+++ b/addons/gut/script_parser.gd
@@ -167,10 +167,8 @@ class ParsedScript:
 		# the right "is_local" flag.
 		if(!is_native):
 			methods = thing.get_script_method_list()
+			methods.reverse()
 			for m in methods:
-				if _methods_by_name.has(m.name):
-					continue
-
 				var parsed_method = ParsedMethod.new(m)
 				parsed_method.is_local = true
 				_methods_by_name[m.name] = parsed_method
--- a/addons/gut/script_parser.gd
+++ b/addons/gut/script_parser.gd
@@ -168,6 +168,9 @@ class ParsedScript:
 		if(!is_native):
 			methods = thing.get_script_method_list()
 			for m in methods:
+				if _methods_by_name.has(m.name):
+					continue
+
 				var parsed_method = ParsedMethod.new(m)
 				parsed_method.is_local = true
 				_methods_by_name[m.name] = parsed_method
--- a/test/integration/test_abstract_class_doubling.gd
+++ b/test/integration/test_abstract_class_doubling.gd
@@ -63,7 +63,7 @@ func test_can_stub_to_return_for_abstract_method_at_instance_level():
 
 func test_error_when_stubbing_to_call_super_at_script_level():
 	stub(AbstractClass, 'abstract_method').to_call_super()
-	pass_test("no errors")
+	assert_tracked_gut_error(gut, 0)
 
 
 func test_error_when_stubbing_to_call_super_at_instance_level():
--- a/test/integration/test_abstract_class_doubling.gd
+++ b/test/integration/test_abstract_class_doubling.gd
@@ -76,8 +76,9 @@ func test_error_when_stubbing_to_call_super_at_instance_level():
 
 	# Assert
 	assert_null(result)
+	var current_test_errors = gut.error_tracker.get_current_test_errors()
+	assert_eq(current_test_errors[0].code, "Cannot call super() because method abstract_method is abstract.")
 	assert_tracked_gut_error()
-	assert_eq(get_logger().get_errors()[0], "Cannot call super() because method abstract_method is abstract.")
 
 
 func test_can_stub_implemented_abstract_to_call_super():
--- a/test/integration/test_abstract_class_doubling.gd
+++ b/test/integration/test_abstract_class_doubling.gd
@@ -1,0 +1,70 @@
+extends GutInternalTester
+
+
+@abstract
+class AbstractClass:
+
+	@abstract func abstract_method()
+
+
+@abstract
+class ExtendsAbstractClass:
+	extends AbstractClass
+
+	func abstract_method():
+		return "implemented"
+
+	@abstract
+	func another_abstract_method()
+
+
+func before_all():
+	register_inner_classes(get_script())
+
+func test_can_double_abstract():
+	var dbl = double(AbstractClass)
+	assert_not_null(dbl)
+
+
+func test_can_stub_to_return_for_abstract_method_at_sctipt_level():
+	stub(AbstractClass, 'abstract_method').to_return('a')
+	var inst = double(AbstractClass).new()
+	assert_eq(inst.abstract_method(), 'a')
+
+
+func test_can_stub_to_return_for_abstract_method_at_double_level():
+	var Dbl = double(AbstractClass)
+	stub(Dbl, 'abstract_method').to_return(9)
+	var inst = Dbl.new()
+	assert_eq(inst.abstract_method(), 9)
+
+
+func test_can_stub_to_return_for_abstract_method_at_instance_level():
+	var inst = double(AbstractClass).new()
+	stub(inst.abstract_method).to_return(7)
+	assert_eq(inst.abstract_method(), 7)
+
+
+func test_error_when_stubbing_to_call_super_at_script_level():
+	stub(AbstractClass, 'abstract_method').to_call_super()
+	pass_test("no errors")
+
+
+func test_error_when_stubbing_to_call_super_at_instance_level():
+	# Arrange
+	var doubled = autofree(double(AbstractClass).new())
+	stub(doubled.abstract_method).to_call_super()
+
+	# Act
+	var result = doubled.abstract_method()
+
+	# Assert
+	assert_null(result)
+	assert_tracked_gut_error()
+	assert_eq(get_logger().get_errors()[0], "Cannot call super() because method abstract_method is abstract.")
+
+
+func test_can_stub_implemented_abstract_to_call_super():
+	var inst = double(ExtendsAbstractClass).new()
+	stub(inst.abstract_method).to_call_super()
+	assert_eq(inst.abstract_method(), 'implemented')
--- a/test/integration/test_abstract_class_doubling.gd
+++ b/test/integration/test_abstract_class_doubling.gd
@@ -21,6 +21,22 @@ class ExtendsAbstractClass:
 func before_all():
 	register_inner_classes(get_script())
 
+
+# Our implementaiton assumes that the order of get_script_method_list() is
+# from child to parents. The following test will show when future godot
+# versions change the behavior.
+func test_method_order_assumption():
+	var method_list = (ExtendsAbstractClass as Variant).get_script_method_list()
+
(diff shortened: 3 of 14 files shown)


bitwes:
It looks like I started checking out your changes too early.  I'm going to leave this here for now and wait until you have any questions or this leaves "draft" status to review anymore.

Here's some thoughts I had
```
"""
Maybe stub_params should make a list of abstract methods in _init.  It should
use is_abstract() to determine if it needs to make the list.

I'm not sure if there are any Native classes that have abstract methods.  I
would assume there are some, so we might have to check get_method_list in
addition to get_script_method_list.

I'm not sure we can assume the ordering of methods in method list functions.

I'm not sure we can assume that only abstract methods can have two listings
for the same method in method list functions.  I've never seen it before, but
there's talk of "interfaces" coming to GDScript, which could lead to multiple
listings for non-abstract methods.

Both of these assumptions might be true now, but I don't think we can assume
that this won't change in the engine later.  If it does, it will likely cause
a hard to track down error.  This is an assumption about assumptions.
"""
```

Here's some tests that I threw together to spot check what you had.  I think these are a good start to some integration tests. Please add these as `test/integration/test_abstract_class_doubling.gd`.  Add any other tests you think of to this script as well.
```gdscript
extends GutInternalTester

@abstract
class AbstractClass:

	@abstract func abstract_method()


@abstract
class ExtendsAbstract:
	extends AbstractClass

	func abstract_method():
		return "implemented"

	@abstract
	func another_abstract_method()


func before_all():
	register_inner_classes(get_script())


func test_can_double_abstract():
	var dbl = double(AbstractClass)
	assert_not_null(dbl)


func test_can_stub_to_return_for_abstract_method_at_sctipt_level():
	stub(AbstractClass, 'abstract_method').to_return('a')
	var inst = double(AbstractClass).new()
	assert_eq(inst.abstract_method(), 'a')


func test_can_stub_to_return_for_abstract_method_at_double_level():
	var Dbl = double(AbstractClass)
	stub(Dbl, 'abstract_method').to_return(9)
	var inst = Dbl.new()
	assert_eq(inst.abstract_method(), 9)


func test_can_stub_to_return_for_abstract_method_at_instance_level():
	var inst = double(AbstractClass).new()
	stub(inst.abstract_method).to_return(7)
	assert_eq(inst.abstract_method(), 7)


func test_error_when_stubbing_to_call_super_at_script_level():
	stub(AbstractClass, 'abstract_method').to_call_super()
	pass_test("no errors")


func test_error_when_stubbing_to_call_super_at_double_level():
	var Dbl = double(AbstractClass)
	stub(Dbl, 'abstract_method').to_call_super()
	assert_tracked_gut_error()


func test_error_when_stubbing_to_call_super_at_instance_level():
	var Dbl = double(AbstractClass)
	var inst = Dbl.new()
	stub(inst.abstract_method).to_call_super()
	assert_tracked_gut_error()


func test_can_stub_implemented_abstract_to_call_super():
	var inst = double(ExtendsAbstract).new()
	stub(inst.abstract_method).to_call_super()
	assert_eq(inst.abstract_method(), 'implemented')
```

namelessvoid:
Thanks for your pre-review :D While finishing the implementation to make all tests pass, I had the same doubts about the assumptions. Therefore I'm gonna rework this and go with my first, simpler approach: Doing the error logging in the double and not inside the `StubParams.to_call_super()`. This is less user-friendly, because the error is logged when the method is actually called as opposed to when `to_call_super()` is invoked. However, it seems to be the less complex, more maintainable and also more stable approach.

bitwes:
Ok.  Speaking of the double error, I did have a comment that got scrapped when I abandoned my review.  I think the `assert` could be changed to `__gutdbl_values.gut.get_logger().error(...)` instead of `assert`.  If for some reason that doesn't work, I think a `push_error` would be better than `assert(false`.  Good luck.  Looking forward to what you come up with.

bitwes:
I asked a question in the godot-engine/gdscript-discuss discord channel.  The response I got indicates that we probably don't have to worry about abstract methods in `get_method_list`.  One safeguard we could add is that if a script is abstract but we can't find any abstract methods in `get_script_method_list` we could either error, warn, or check `get_method_list`.  I don't think we need to check it by default.

namelessvoid:
@bitwes I drafted a somewhat simpler approach, which keeps changes "local" to the method maker. So we have to do the magic only in one place instead of two.

test/integration/test_abstract_class_doubling.gd
```diff
@@ -0,0 +1,70 @@
...
+
+func test_can_stub_to_return_for_abstract_method_at_instance_level():
+	var inst = double(AbstractClass).new()
+	stub(inst.abstract_method).to_return(7)
+	assert_eq(inst.abstract_method(), 7)
+
+
+func test_error_when_stubbing_to_call_super_at_script_level():
+	stub(AbstractClass, 'abstract_method').to_call_super()
+	pass_test("no errors")
```
namelessvoid, line 50: Not sure how to turn this into a meaningful test :)

bitwes:
I hate to do this, but I'm starting to think we are going about this the wrong way.  After I saw your changes to `script_parser.gd` I started looking around and I'm doing a lot of the same thing in multiple places and in different ways.  I think it might be time to consolidate some of this into a single place.  Take a look at this idea and let me know what you think.

# script_parser.gd approach
The premise is to add the abstract method detection logic to `script_parser.gd` and then have a static/global instance that is used wherever it is needed.  Right now, if we wanted to use it in `stub_params` then every instance of `stub_params` would have its own copy of a script parser and we'd be parsing scripts each a stub is created.

I don't think there's an easy way to pass around an instance of `script_parser` between doubler/stubber/stub_params/test without adding one to `gut.gd` and at that point it might as well be static/global.  

By having it static/global, we aren't doing any extra work since anything that gets a stub has to be used on a double and making a double requires parsing the script.  It may change the timing of when a script is parsed for the first time, but that doesn't matter.

The easiest way I see to do this is to change `script_parser.gd.scripts` to be `static`.  This means we won't have to manage  change any other existing code.  It might generate some warnings if there are any direct references to `scripts` since they will be going through an instance instead of the class.  

The right thing to do would probably be to make a global by adding a static `var` to `utils.gd` (call it `script_collector`).  If we do this, then everything that needs a `script_collector` needs a local variable that defaults to `GutUtils.script_collector`.  These will then use the local variable internally.  This gives us the ability to inject a different value for testing purposes.

## Changes
1.  Do whatever needs to be done for static/global approach.
1.  Add logic to `script_parser.gd.ParsedScript` to detect abstract methods (more on that later).
1.  Add a bool to `script_parser.gd.ParsedScript` for the result of `is_abstract()`.
1.  In `stub_params.gd` use the new static/singleton/global `script_parser.gd` to check if a method is abstract or not.
1.  In `doubler.gd` use the new static/singleton/global `script_parser.gd` to check if a method is abstract or not.


# Consolidate other logic (optional)
This is a place where any other logic we find in this process can be documented.  If we don't address anything in this list I can make a "technical debt" issue to track it and address it when I get bored.

## Getting reference to native classes by name.
1.  Move `stubber.gd._make_crazy_dynamic_over_engineered_class_db_hash` to utils.gd in a new `static_init` and store results in a static var.  Let's call it `native_classes_by_name`
1.  Use this in script_parser.gd._get_native_methods()
1.  In `stubber.gd` replace `_class_db_name_hash` with `GutUtils.native_classes_by_name`

bitwes:
# Detecting abstract methods
Here's a snippet of code that might be able to be used to get a list of all normal and abstract methods in a script.  I haven't thought about how to use it in `script_parser.gd` but I think this works.
```gdscript
func parse_methods(klass):
	var normal_methods = {}
	var abstract_methods = {}
	for method in klass.get_script_method_list():
		if(method.flags & METHOD_FLAG_VIRTUAL_REQUIRED):
			abstract_methods[method.name] = method
		else:
			normal_methods[method.name] = method

	# Adds an "implmeted" field to the meta.  I don't think we need it, but this
	# is one way you could tell.
	for key in normal_methods:
		if(abstract_methods.has(key)):
			abstract_methods.erase(key)
			normal_methods[key].implemented = true
		else:
			normal_methods[key].implemented = false

	return {
		"normal":normal_methods,
		"abstract":abstract_methods
	}
```
You can see it in action here:  https://github.com/bitwes/Gut/blob/i761_abstract_classes/scratch/abstract_classes.gd

namelessvoid:
Thanks for sharing your ideas!

I had a look at your suggestion and I agree: Consolidating the logic in one place makes sense. However, I'm not sure if the `ScriptParser` is the right place. Right now, I don't have enough knowledge about the code-base yet to see if this is the optimal approach.

In addition, I think there might be also a win in removing some indirections. For example, a doubled class delegates a lot of logic e.g. to the stubber, but on the other hand also does a lot of logic by itself. To me it feels like the separation of concerns is not cut optimally in this case. When this is tackled, it might solve some of the redundancies already.

That being said, I see potential of improvement, but I'd suggest to do that in a separate PR. For the time being, my implementation works as expected and merging would close #763 quickly.

addons/gut/script_parser.gd
```diff
@@ -168,6 +168,9 @@ class ParsedScript:
 		if(!is_native):
 			methods = thing.get_script_method_list()
 			for m in methods:
+				if _methods_by_name.has(m.name):
+					continue
```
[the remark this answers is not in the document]
bitwes: I think this line is causing most of the failures in other tests.

test/integration/test_abstract_class_doubling.gd
bitwes: replace `pass_test` with `assert_tracked_gut_error(gut, 0)`.  Now that we know what kind of error is thrown we can test for it.

bitwes:
> For the time being, my implementation works as expected and merging would close #763 quickly.

If you can resolve the failing tests in a full run, then I may consider merging this.  I'm not too concerned with closing this issue out quickly.  There is a decent workaround and I would much rather have a solid solution with the least amount of technical debt.

The `script_parser` approach might simplify addressing the failing tests and may make it easier to make `stub_params` error at the time you call `to_call_super` as opposed to when the method on the double is called.

> For example, a doubled class delegates a lot of logic e.g. to the stubber, but on the other hand also does a lot of logic by itself.

Are you talking about the actual code generated for a doubled class vs the stubber?  I'm interested in hearing more if you want to elaborate.

namelessvoid:
Oh dear, I ran the tests and thought only UI interactions are failing (which I thought is related to https://github.com/bitwes/Gut/issues/746). I will re-run and fix them, of course!

But in the end you're right - there is a workaround and right now I seem to be the only one having issues. And I'm already using my patch in my project, so no hurries!

> Are you talking about the actual code generated for a doubled class vs the stubber? I'm interested in hearing more if you want to elaborate.

Yes, I'm talking about that. I'd need some time to investigate before I post something random, but I found it confusing that the double does some calls to stubber and then to do some if/else to decide which stubber method to call eventually. That might go as well into a single class - maybe even into the stub params. And another high-level idea: if we have a `ParsedScript` object with meta data about a doubled class, we might set a reference to that on the double right away. So we could move logic from the stubber to the double (I mean `double_tools`).

But as I said, this is more high-level brainstorming without knowing all the details :)

namelessvoid:
Alright, tests are fixed. I also fixed the log line assert in the abstract class test - it was failing when all tests have been run.

test/resources/doubler_test_objects/double_extends_node2d.gd
bitwes, line 1: Revert these changes.

test/unit/test_doubler.gd
bitwes, line 1: Revert these changes.

test/integration/test_doubler_and_stubber.gd
bitwes, line 1: Revert these changes.

addons/gut/doubler.gd
bitwes, line 1: Revert these changes.

namelessvoid:
Unrelated changes have been reverted, feel free to have another look!

PR #765, opened 2025-09-22
fix error calling get_base_script on null instance for non-script files

### LINKED ISSUE #764: .cfg files opened in editor cause error: Attempt to call function 'get_base_script' in base 'null instance' on a null instance
# Versions

* Godot: v4.5.stable.official [876b29033]
* GUT: 9.5.0
* OS: MacOS 26.0

# The Bug
I'm running godot 4.5 with GUT 9.5.0 and get these errors when launching the project:
(just the editor, VS Code NOT opened)

```
Add Autoload
--- GDScript language server started on port 6005 ---
  ERROR: res://addons/gut/editor_caret_context_notifier.gd:176 - Attempt to call function 'get_base_script' in base 'null instance' on a null instance.
  ERROR: res://addons/gut/editor_caret_context_notifier.gd:176 - Attempt to call function 'get_base_script' in base 'null instance' on a null instance.
```

when adding some logging I can see that _on_script_changed is called for a bunch of scripts and two null instances:
```
<Object#null>[GDScript backtrace (most recent call first):
    [0] is_test_script (res://addons/gut/editor_caret_context_notifier.gd:179)
    [1] _on_editor_script_changed (res://addons/gut/editor_caret_context_notifier.gd:145)]
  ERROR: res://addons/gut/editor_caret_context_notifier.gd:180 - Attempt to call function 'get_base_script' in base 'null instance' on a null instance.
```
(it is already null in _on_editor_script_changed)

I noticed closing all script tabs in the editor was resolving this.

when opening things back up again this error returned when I opened some plugin.cfg files inside the in Editor Code Editor!
Built in docs do not lead to these errors, so seems like .cfg files (or maybe all unsupported file types that can be opened in the code editor?) cause this.

# Steps To Reproduce
- open a plugin.cfg file inside the godot editor code editor (you will see the error here already)
- reload the project
- see errors in console

Edit: .json files also cause this error

ThunderFD:
fixes #764

Opening non-script files within the editor calls _on_editor_script_changed with null as its argument, leading to some errors printed to console when e.g. opening json files or if such files are open already when starting the editor.

as I'm not familiar with the code base please check if more values need to be reset when null is passed here (or if this is already sufficient).

thanks!

### DIFF
--- a/addons/gut/editor_caret_context_notifier.gd
+++ b/addons/gut/editor_caret_context_notifier.gd
@@ -124,12 +124,17 @@ func _make_info(editor, script, test_script_flag):
 # Events
 # -------------
 
-# Fired whenever the script changes.  This does not fire if you select something
-# other than a script from the tree.  So if you click a help file and then
-# back to the same file, then this will fire for the same script
+# Fired whenever the script changes.  This does not fire for help files.  If
+# you click a help file and then back to the same file, then this will fire
+# for the same script
+#
+# This does fire for some non-script files such as .cfg, .json and .md files,
+# but the passed in value will be null.
 #
 # This can fire multiple times for the same script when a script is opened.
 func _on_editor_script_changed(script):
+	if(script == null):
+		return
 	_last_line = -1
 	_current_script = script
 	_current_editor_base = _current_script_editor.get_current_editor()


addons/gut/editor_caret_context_notifier.gd
```diff
@@ -130,6 +130,12 @@ func _make_info(editor, script, test_script_flag):
 #
 # This can fire multiple times for the same script when a script is opened.
```
bitwes, line 131: Please update this comment block to the following:
```
# Fired whenever the script changes.  This does not fire for help files.  If
# you click a help file and then back to the same file, then this will fire
# for the same script
#
# This does fire for some non-script files such as .cfg and .json files, but
# the passed in value will be null.
#
# This can fire multiple times for the same script when a script is opened.
```

bitwes (commented):
I tested out the changes and it looks good.  Please make the change to the comment block above the method and I can merge this.  Thanks for the PR.

bitwes:
Actually, no values need to be reset.  It should just return early.  I started thinking about it, and everything works fine when you click a help file and the signal is not emitted, so if we just return and not set the values, then it's like it wasn't called.

PR #769, opened 2025-09-28
677 elapsed time

### LINKED ISSUE #677: Track elapsed time for a test and make it accessible to tests.
# Versions
Gut 9

# The Feature
GUT should make the execution time for a test available to the test.  This way you can make assertions that something has occurred in an expected amount of time.

Use case:
* Verifying spawn rate.  In some cases it might be easier to wait for X number of enemies to spawn and then check the count rather than awaiting a specified amount of time.

Functions/Properties
* `get_elapsed_time()` method
* `assert_elapsed_time_eq` (probably not that useful).
* `assert_elapsed_time_almost_eq` (maybe the best one)
* `assert_eplased_time_lt` (probably the 2nd most useful).
* `asssert_elapsed_time_gt` (probably not that great.

Curtis-Barnhart:
resolves #677
By adding a variable to GutTest that can be set to the current time by a GutMain instance before a test method is called, a test can track how much time has elapsed since that method began. We expose `get_elapsed_time`, which allows the user to obtain the amount of time elapsed since the test started, as well as four assertion methods, `assert_elapsed_time_eq`, `assert_elapsed_time_almost_eq`, `assert_eplased_time_lt`, and `asssert_elapsed_time_gt` that allow the user to make assertions about the amount of time passed.


addons/gut/test.gd
```diff
@@ -791,6 +794,11 @@ func compare_deep(v1, v2, max_differences=null):
 	return result
 
 
+## Returns the time that the test method has been running in seconds as a float.
+func get_elapsed_time() -> float:
+	return Time.get_unix_time_from_system() - self._test_began_at
```
bitwes, line 799: Remove `self`.

addons/gut/test.gd
```diff
@@ -791,6 +794,11 @@ func compare_deep(v1, v2, max_differences=null):
 	return result
 
 
+## Returns the time that the test method has been running in seconds as a float.
+func get_elapsed_time() -> float:
```
bitwes, line 798: Rename to `get_elapsed_seconds`.

bitwes (changes requested):
Thank you.  I almost implemented this last week because I needed it for my game.  

I think I would like to eventually add `get_elapsed_msecs`, `get_elapsed_usecs`, and maybe `get_elapsed_idle_frames`, `get_elapsed_physics_frames`.  I think adding extra asserts for each of these would be too many special asserts.  For that reason I would like you to remove the new asserts you added.  

If you want to implement any of the other get methods that's great, but if you just remove the asserts and regenerate the help files, I'll merge this.

bitwes (changes requested):
These are going to be great additions.  I have some implementation and naming changes.  As soon as I started reviewing I realized I should have described what I was thinking earlier.  Sorry about that.

Curtis-Barnhart:
I've made all the changes you requested (and as one additional thing, renamed get_elapsed_seconds to get_elapsed_sec so it is more in line with the other elapsed time methods).

bitwes:
This works, and looks pretty good, but that's a lot of private vars being accessed by `gut.gd`.  I may have missed that in the earlier reviews, but now that there is 6 in a row it really stands out.

Add `test.gd.reset_start_times` and call that from `gut.gd` and we'll be good to go!
```gdscript
func reset_start_times():
	_unixtime_began_tracking = Time.get_unix_time_from_system()
	_elapsed_msec_start = Time.get_ticks_msec()
	_elapsed_usec_start = Time.get_ticks_usec()
	_elapsed_physics_frames_start = Engine.get_physics_frames()
	_elapsed_process_frames_start = Engine.get_process_frames()
```

Curtis-Barnhart:
All done.

PR #774, opened 2025-10-02
Working

AdminCrystal:


bitwes:
This is a bit of a gross problem due to how nested `diff_tool` is.  The majority of values in `gut_config.gd` are applied directly to `gut.gd`.  This is done in `gut_config.gd._apply_options`.  In some cases `gut.gd` just holds values that `test.gd` will use via its `gut` variable, when the test is run.  I think this is probably how we would have to do this.  Then we'll pass the value along to `comparator.gd` which will use it internally and pass it along to `get_short_summary`.

I see two places where output is being truncated.  Wherever `comparator.gd` is using its `_max_length` (which is 100) property and `diff_tool.gd.get_short_summary()` (which is hardcoded to 50).  `get_short_summary` could have a default parameter that it would use and `comparator.gd` could pass in `_max_length` (which we should rename).

After a quick look, I think this would work.  I've chosen `truncate_value_length` (over `max_console_output_length`).  I think it describes how it is used, especially after it is added in the multiple different places.  If you come up with something better, you could use that.

* `diff_tool.gd`
    * Give `get_short_summary` a default paramter that it will use instead of the hardcoded `50`.
    ```
    func get_short_summary(truncate_longer_than=50)
    ```

* `comparator.gd`
    * Change `_max_length` to `truncate_value_length`
    * Pass `truncate_value_length` to all calls of `get_short_summary`.
    

* `gut.gd`
    * Add a property called `truncate_value_length`.  Default it to 100.

    
* `test.gd`
    * In `_do_ready_stuff` (I agree, horrible name, shouldn't be there, see comment) add
    ```
    _compare.truncate_value_length = gut.truncate_value_length
    ```


* `gut_config.gd`
    * Change `max_console_output_length` to `truncate_value_length`
    * In `_apply_options` add
    ```
    gut.truncate_value_length = opts.truncate_value_length
    ```

I don't know why I chose 50 for `diff_tool` and 100 for `comparator.gd`.  This change would make the default for both to be 100.  That might be too much output, but it is probably fine.  I'd have to see it in action.

bitwes:

PR #776, opened 2025-10-03
Return Params start

AdminCrystal:


AdminCrystal:
As mentioned in our other pull request I have made this one as just the return types for public methods, I avoided any @deprecated @ignore or ones prefixed with _ for this pr. 
I tried to put specific return types where possible, theoretically some of the ones I labeled as Variant could be specified as Node instead but not sure if that would break something
Some could be improved such as compare_deep instead of returning Variant it could return the CompareResult but since its not an actual class its not possible to do that with current set up

I did run the tests before and after my changes but I have failures even on main branch pre my changes, they haven't changed counts but I thought I should mention before you think about merging 

Godot v4.5.stable - macOS Sequoia (15.6.1) - Multi-window, 1 monitor - Metal (Forward+) - integrated Apple M4 Pro (Apple9) - Apple M4 Pro (12 threads) - 24.00 GiB memory
Screenshot 2025-10-02 at 8 10 03 PM

addons/gut/spy.gd
```diff
@@ -62,7 +62,7 @@ func was_called(variant, method_name, parameters=null):
 	return to_return
 
 
-func get_call_parameters(variant, method_name, index=-1):
+func get_call_parameters(variant, method_name, index=-1) -> Variant:
```
bitwes, line 65: `Array`, optionally `Array[Variant]`, I'm fine with either, we just have to be consistent.

bitwes:
Please post your failing tests.  My prediction is that they are input related.

addons/gut/test.gd
```diff
@@ -778,13 +778,13 @@ func skip_if_godot_version_ne(expected):
...
 ## before you can double any inner class.
-func register_inner_classes(base_script):
+func register_inner_classes(base_script) -> void:
 	gut.get_doubler().inner_class_registry.register(base_script)
 
 
 ## Peforms a deep compare on both values, a CompareResult instnace is returned.
 ## The optional max_differences paramter sets the max_differences to be displayed.
-func compare_deep(v1, v2, max_differences=null):
+func compare_deep(v1, v2, max_differences=null) -> Variant:
```
bitwes, line 787: `Array`

addons/gut/test.gd
```diff
@@ -2501,7 +2501,7 @@ func get_summary_text():
 
 ## Create a Double of [param thing].  [param thing] should be a Class, script,
 ## or scene.  See [wiki]Doubles[/wiki]
-func double(thing, double_strat=null, not_used_anymore=null):
+func double(thing, double_strat=null, not_used_anymore=null) -> Variant:
```
bitwes, line 2504: This can be `Object` I think.  Unless you can't return `null` for `Object`, then it has to remain `Variant`.  I always forget where I can/can't use `null` when something has a type.  I think primitives are the only thing that can't be `null`.

addons/gut/test.gd
```diff
@@ -2510,7 +2510,7 @@ func double(thing, double_strat=null, not_used_anymore=null):
 
 ## Create a Partial Double of [param thing].  [param thing] should be a Class,
 ## script, or scene.  See [wiki]Partial-Doubles[/wiki]
-func partial_double(thing, double_strat=null, not_used_anymore=null):
+func partial_double(thing, double_strat=null, not_used_anymore=null) -> Variant:
```
bitwes, line 2513: Same as `double`

bitwes (changes requested):
I added comments for some methods that I think can be more specific than `Variant` but we would have to run the tests to be sure.

bitwes:

PR #793, opened 2025-12-26
Add support for anonymous functions to assert_connected 

### LINKED ISSUE #786: assert_connected does not work with anonymous functions
# Versions
*(list all versions where you have replicated the bug)*
* Godot: 4.5.1
* GUT: 9.5.1
* OS: MacOS

# The Bug
The following test fails:
```gdscript
func test_demo_issue():
	var btn = autofree(Button.new())
	var f = func(): print('hello')
	btn.pressed.connect(f)
	assert_connected(btn.pressed, f)
```

Curtis-Barnhart:
Resolves #786

Previously anonymous functions could not be tested for connection in assert_connected and assert_not_connected. The connection assertions relied on using the names of methods instead of references to them, which gets a little messy with anonymous function because they are all just named "\<anonymous function\>" (or something similar, I don't remember exactly what).

This PR implements a new system for testing signal/method connection that stores references to methods instead of just their names (and adds tests to make sure this behavior is well documented). The signature of the assertion methods was also a little bit complicated and untyped, and so I took the liberty of adding a push_error informing the user they have incorrectly called the assertion method in the case that they have. I can certainly remove this part if it is found unnecessary.

### DIFF
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -1361,51 +1361,55 @@ func _is_connected_to_any(
 	return false
 
 
-# Helper function to parse parameters of assert_[not_]connected
-# Attempts to parse parameters as each of the four signatures
-# of assert_[not_]connected and return whether or not the signal/methods(s)
-# specified by the parameters are actually connected.
-# returns [connected, signal_name, method_name, signal_object_name, method_object_name],
-# where connected is true if the objects are actually connected and false otherwise.
-# The remaining parameters are strings used to display test information.
-func _get_connection_info(p1, p2, p3, p4):
+# Helper return type of _get_connection_info.
+# connected is true if the objects are actually connected and false otherwise.
+# The remaining parameters are used to display test information.
+class _ConnectionInfo:
+	var connected: bool = false
 	var signal_name: String = ""
 	var method_name: String = ""
 	var signal_object_name: String = ""
 	var method_object_name: String = ""
-	var connected: bool = false
+
+
+# Helper function to parse parameters of assert_[not_]connected
+# Attempts to parse parameters as each of the four signatures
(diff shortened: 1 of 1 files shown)


addons/gut/test.gd
```diff
@@ -1422,22 +1470,23 @@ func assert_connected(p1, p2, p3=null, p4=""):
...
-		connect_to_obj = connect_to_obj.get_object()
-
-	var method_disp = ''
+func assert_not_connected(p1, p2, p3=null, p4=null):
+	var conn_result = _get_connection_info(p1, p2, p3, p4)
+	var connected = conn_result[0]
+	var signal_name = conn_result[1]
+	var method_name = conn_result[2]
+	var signal_object_name = conn_result[3]
+	var method_object_name = conn_result[4]
```
bitwes: Same as `assert_connected`.

bitwes (changes requested):
Thanks.  These signal asserts are very unwieldy, and this is a good patch to get the new assert functionality.    

Just a minor change to not replicate the ugliness of `SignalAssertParameters.others`.  Changing `_get_connection_info` to return a dictionary might help refactoring that class later too (it needs it).

Curtis-Barnhart:
Would it work to have the return type just be a class, so the editor can help prevent misspellings/give type hints on top of the benefits of using a dictionary instead of an array? This is what I've implemented, but if you'd prefer a raw dictionary I can change it to that as well.

bitwes:
In my heart, I wanted a class, but it probably wasn't necessary.  A dictionary sounded like a reasonable compromise.  So I'm fine with a class.  I'll merge this in after I get 4.6 merged (later today, I already verified it will merge w/o conflict).

Issue #762, opened 2025-09-19
Global lifecycle hooks


aleperaltabazas:
# Versions

Godot 4+

# The Feature

GUT offers the possibility for each test class that inherits GutTest to redefine `before_each`, `before_all`, `after_each` and `after_all` hooks to execute certain code before or after each event. However, this is limited to each test suite and does not allow to register common behaviour for each event and thus is necessary for state to reset to duplicate code all throughout.

One simple solution to this problem exists, which is editing the GutTest class (or extending it with a custom script) to redefine the hook methods, but this also requires the explicit call of super in every other test, which may be error prone.

I suggest for GUT to have a mechanism to register global hooks in any point and for them to affect _all_ tests, similar to [Kotest's extensions system](https://kotest.io/docs/framework/extensions/extensions-introduction.html).

The following is an example of what the extension could be like:

```gd
# defines methods for the hooks with default, null behaviour for every one of them, user is expected to override whichever method/s they need
class GutExtension
extends RefCounted


# called before every test in every suite
func before_each():
  pass


# called once for every suite, before every suite
func before_suite():
  pass


# called only once before any test is run
func before_any():
  pass


# called after every test in every suite
func after_each():
  pass


# called once for every suite, after every suite
func after_suite():
  pass


# called only once after every test is run
func after_every(): # bad name, expressive but doesn't match with before_any
  pass
```

As for using it, the following is an example use from a pre-run hook script.

```gd
extend GutHookScript


class CustomHook extends GutHook:
  func before_each():
    reset_my_state()

  func after_all():
    tear_down()


func run():
  GutMain.register(CustomHook.new())
```

I'm working on a little PoC on the project I have for it, as I commonly need to reset state to defaults, and I end up having a lot of before and afters with the same code. Would it be fine if I extract it into a fork to show the way it could be used?

bitwes:
> similar to [Kotest's extensions system](https://kotest.io/docs/framework/extensions/extensions-introduction.html).

Interesting.  If nothing else, I think we should lobby Godot to allow us to use `fun` instead of `func`.  `override suspend fun`...does that mean you are preventing fun from being suspended, as to ensure the good times continue?

> Would it be fine if I extract it into a fork to show the way it could be used?

Absolutely.  I can see the need for more hooks/setup/teardown.  I haven't run into any cases yet myself, but I think I created an issue about adding more hooks (at least a pre-script and post-script hook).  

> or extending it with a custom script

[...]

PR #794, opened 2025-12-27
Added methods to modify Engine.time_scale for the duration of a single test

### LINKED ISSUE #401: No mentions of `Engine.set_time_scale`
I've found the function extremely useful in speeding up my testing dramatically, but it doesn't seem to be mentioned anywhere in the repo.  

Everyone should be using this.

Selection_061

Curtis-Barnhart:
Resolves #401

Issue 401 describes how the time_scale member of Engine can be changed to significantly speed up some types of tests (it can cause timing issues with others, so it should be used carefully). bitwes mentioned there that it would be good to have something built in to gut that could reset this value after tests were run so unexpected Engine time scales wouldn't bleed into other code.

This PR adds a member and two methods to the GutTest class that allow users to easily modify the Engine time scale for individual tests inside that test's code without needing to worry about cleaning up values. To change the time scale of any given test:

```gdscript
class ExampleTest:
    extends GutTest

    func test_something():
        set_time_scale(5.0)
        await wait_seconds(5.0)
        # By now, about 1 second has passed
```

This `set_time_scale` method can also be called inside before_each so that every method in a test runs on a given time scale. The time scale can also be set to multiple different values multiple times within a single test method, and after the method ends, it will always be reset to the original Engine value before the first set_time_scale was called. The other method is `time_scale_reset` which resets the time scale set through `set_time_scale` and can be called anytime during the test method by the user as well.

I will also regenerate the class reference as well as update the wiki documentation if this approach in the code is to your liking - I didn't want to write up a wiki description of how it worked until I knew that the approach was actually final, haha. Does this seem like an acceptable solution to the ideas proposed in 401?

### DIFF
--- a/test/unit/test_test.gd
+++ b/test/unit/test_test.gd
@@ -1791,7 +1791,7 @@ class TestTimeScale:
 
 
 # ------------------------------------------------------------------------------
-class TestTimeScaleBeforeAll:
+class TestTimeScaleBeforeEach:
 	extends BaseTestClass
 
 	func before_each():


bitwes:
I like the idea, but in reading up on `time_scale`, I think it has too many caveats to manage this in GUT.  As soon as we add a wrapper, users tend to think "you let me do this, why aren't you protecting me from myself?"  I'd want to know a ton more about how things can go wrong before making this a feature of GUT.

Maybe a Tips/Tricks page would be a good place to start.  We could put in an example `FastTest` class with a lot of warnings and links to Godot docs.  Maybe a call-to-action to give feedback if they have any success with this approach.
``` gdscript
class_name MyFastTest
extends GutTest
var default_time_scale := Engine.get_time_scale() # or just 1.0
# add in physics and maybe audio ticks as well.
var fast_scale := 2.0

func before_each():
   # Having a way to disable fast tests is probably a good idea.
   if(MyGlobalsSomewhere.use_fast_tests):   
      Engine.set_time_scale(fast_ticks)

func after_each():
   Engine.set_time_scale(default_time_scale)

func after_all():
   Engine.set_time_scale(default_time_scale)

```
```gdscript
extends MyFastTest

func before_all():
    fast_scale = 4.5

func before_each():
   super()
   ...
func after_each():
   super()
   ...
func after_all():
   super()
   ...

<example tests>
```

This might also be good place to reference the hooks [described here](https://github.com/bitwes/Gut/issues/762#issuecomment-3315138513) (but not yet added to the documentation).  Being able to also reset the scales after a script has finished running (just in case something goes really wrong in a test script) would be good.

I'm most concerned about what it looks like to keep `time_scale` and `physics_ticks_per_second` synced to reasonable values across different platforms/environments.  Especially when CI/CD pipelines are involved.  This could all work fine locally but the pipeline becomes very flakey.

> Note: It's recommended to keep this property above 0.0, as the game may behave unexpectedly otherwise.

> Note: This does not affect audio playback speed. Use AudioServer.playback_speed_scale to adjust audio playback speed independently of Engine.time_scale.

> Note: This does not automatically adjust physics_ticks_per_second. With values above 1.0 physics simulation may become less precise, as each physics tick will stretch over a larger period of engine time. If you're modifying Engine.time_scale to speed up simulation by a large factor, consider also increasing physics_ticks_per_second to make the simulation more reliable.

Curtis-Barnhart:
Okay, it makes sense to not officially add it to this codebase since messing with the time scale does seem a little bit sketchy. Since it seems like what would be most helpful in this case is more documentation and a pivot away from the original goal of this PR (especially since you mention other useful documentation like the hooks from that other issue), I'm going to close this PR and open a new one specifically for documentation.

Issue #788, opened 2025-12-06
"exclude_addons" settings removed in 4.6-dev6: Trying to assign value of type 'Nil' to a variable of type 'bool' on gut_loader.gd line 35


isaaccp:
# Versions
*(list all versions where you have replicated the bug)*
* Godot: Godot 4.6-dev6
* GUT: 9.5.0
* OS: Windows 11

Since I updated to Godot 4.6-dev6, I am getting the following error when trying to run GuT:
Trying to assign value of type 'Nil' to a variable of type 'bool'.
on
https://github.com/bitwes/Gut/blob/7ed501f99d5879601b730ffa1d48d061eb2a6ba8/addons/gut/gut_loader.gd#L35

Looking at the ProjectSettings doc at HEAD I can see that the exclude_addons setting has been removed:
https://docs.godotengine.org/en/latest/classes/class_projectsettings.html
whereas it is there in 4.5:
https://docs.godotengine.org/en/4.5/classes/class_projectsettings.html

bitwes:
I saw this yesterday but haven't fixed it yet.  It looks like they changed the settings so you can now configure any directory to be excluded/included.  

Until I fix this on the 4_6 branch of GUT, you can set the variable were_addons_disabled to false and remove the rest (in the file you linked).

bitwes:
This has been fixed on the [4_6](https://github.com/bitwes/Gut/tree/godot_4_6) branch.  You can download it here:  https://github.com/bitwes/Gut/archive/refs/heads/godot_4_6.zip

This will remain open until Godot 4.6 is released and the 4_6 branch is merged into main.

isaaccp:
Thanks!

kyrosle:
4.6 has released and i meet this problem

bitwes:
I haven't merged the 4.6 changes into main yet, and won't be updating the version in the Asset Library until 4.6 has been out for awhile.  You can download the version for 4.6 here https://github.com/bitwes/Gut/archive/refs/heads/godot_4_6.zip.  Unzip it and put the gut folder in your addons folder.

bitwes:
Everything has been merged and Gut 9.6.0 for Godot 4.6 has been released.

PR #813, opened 2026-02-26
Added maximum version for v9.4.0

TheDuckCow:
I would not be heartbroken if you felt this change was not worth committing, but would have saved me at least a little bit of time. What I tested:

- 9.4.0 -> Works in Godot 4.3, 4.4, and 4.5; explicit errors in 4.6
- 9.5.0 -> Does not load in 4.4 (unsurprising), didn't test in 4.5 as I presumed this working, and does not work in 4.6
- 9.6.0 -> Does not load in 4.4 (unsurprising)

## Big caveat

Edit: I thought of this only after posting the PR, but worth saying: I did not do *exhaustive* testing to ensure *all* aspects of gut are working in 9.4.0 with 4.4 and 4.5. I only found the tests that I already implemented in my repo continued to work without modification. Maybe more testing would be required. Again, no feelings hurt if you close it.

### Context

I'm a plugin developer, and am just now upping the minimum version of the [Godot Road Generator](https://github.com/TheDuckCow/godot-road-generator) to 4.4+, now actively supporting 4.4 - 4.6. For a couple of reasons, awhile go I decided to directly commit gut into repo directly. This was to reduce friction with users contributing (fewer setup steps, since there's no _canonical_ plugin version manager out there), as well as resolving issues I had in the past trying to install gut just-in-time for github action testing (though maybe I'd have more luck if I attempted this again).

I was fingers-tapping-together hoping I could use the 9.4.0 version for all 4.4+, but alas it does not work in 4.6 for reasons I'm sure are out of your control. Anyhow, that's my problem to bear, but the doc change might be nice. I suppose it could be helpful to "cap" any other versions that have maximum support levels, as the table header otherwise implies they work in any version above the listed target.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ You can download/clone earlier versions of GUT using the links below.
 |-|-|-|-|-|-|
 |Main Branch |4.6 (or greater)| |[download](https://github.com/bitwes/Gut/archive/refs/heads/main.zip) | | [wiki](https://gut.readthedocs.io/en/latest/)|
 |[9.6.0](https://github.com/bitwes/Gut/releases/tag/v9.6.0)|4.6|[repo](https://github.com/bitwes/Gut/tree/v9.6.0)|[download](https://github.com/bitwes/Gut/archive/refs/tags/v9.6.0.zip)|[Asset Library](https://godotengine.org/asset-library/asset/1709)|[wiki](https://gut.readthedocs.io/en/v9.6.0)|
-|[9.5.0](https://github.com/bitwes/Gut/releases/tag/v9.5.0)|4.5|[repo](https://github.com/bitwes/Gut/tree/v9.5.0)|[download](https://github.com/bitwes/Gut/archive/refs/tags/v9.5.0.zip)||[wiki](https://gut.readthedocs.io/en/v9.5.0/)|
+|[9.5.0](https://github.com/bitwes/Gut/releases/tag/v9.5.0)|4.5 (only)|[repo](https://github.com/bitwes/Gut/tree/v9.5.0)|[download](https://github.com/bitwes/Gut/archive/refs/tags/v9.5.0.zip)||[wiki](https://gut.readthedocs.io/en/v9.5.0/)|
 |[9.4.0](https://github.com/bitwes/Gut/releases/tag/v9.4.0)|4.3-4.5|[repo](https://github.com/bitwes/Gut/tree/v9.4.0)|[download](https://github.com/bitwes/Gut/archive/refs/tags/v9.4.0.zip)||[wiki](https://gut.readthedocs.io/en/v9.4.0/)|
 |[9.3.0](https://github.com/bitwes/Gut/releases/tag/v9.3.0)|4.2|[repo](https://github.com/bitwes/Gut/tree/v9.3.0)|[download](https://github.com/bitwes/Gut/archive/refs/tags/v9.3.0.zip)||[wiki](https://gut.readthedocs.io/en/9.3.1/)|
 |[9.1.1](https://github.com/bitwes/Gut/releases/tag/v9.1.1)|4.1|[repo](https://github.com/bitwes/Gut/tree/v9.1.1)|[download](https://github.com/bitwes/Gut/archive/refs/tags/v9.1.1.zip)|||


bitwes:
I checked out your plugin and it looks impressive.  Best of luck with your game too!  I like hearing from other plugin developers and how they use GUT.  `@tool` scripts and editor functionality can be a pain to test.  Don't hesitate to provide any other feedback or issues.

# What about?
What if the heading was changed to "Valid Godot Versions" and the versions of Godot were listed like this:
9.6.0 | 4.6+
9.5.0 | 4.5.x
9.4.0 | 4.3.x - 4.4.x (why not 4.5.x is covered below)
9.3.0 | 4.2.x
...

I'd like to avoid putting "(only)" on all the GUT versions that only support one minor version of Godot.  I've struggled with keeping that table (and surrounding documentation) as succinct as possible (I have a problem with that...as you can see by the size of this reply and the length of this sub-comment I have put into parenthesis).

I'm currently working on a "check for updates" feature and better Godot/GUT version validating logic.  Currently GUT has a minimum required version hardcoded and that's all the checking it does.  GUT can't check for an upper bound since that can't be known at the time of release.  This new feature would allow me to update that info externally and give better info.

# Godot required version bumps
I don't like to bump the required Godot version unless I have to.  Each time a bump occurs it is because Godot introduced something that broke GUT or I have started using a new feature of Godot (usually both).  I haven't done the best job of documenting why the Godot version was bumped though.  I think this should get added to the release notes going forward.  For now, I added it below.


## Godot 4.5 
Godot introduced their own `Logger` class which clashes with GUT's `Logger` class.  I think this is the only reason why GUT 9.4.0 would not work with Godot 4.5.  I don't think there is an actual issue, just that you get a lot of warnings that GUT's Logger is shadowing Godot's Logger.  There might be problems with trying to use Godot's Logger from within a test, but I'm not sure.

GUT 9.5 introduced new features that won't work in Godot 4.4.  Most of these have to do with asserting that errors or warnings have been generated and failing tests when errors occur (all of this is possible b/c of Godot's new `Logger` class).  Making these kinds of assertions are great for plugins.  You can make assertions against your calls to `push_error` and `push_warning`.  I think this release also gives much better orphan information.

## Godot 4.6
I think #788 is the only issue that prevents GUT 9.5.0  from working in 4.6.  This might be the only change that GUT 9.4 would need (in addition to 4.5 change) to work in 4.6.  You'd have to do some conditional logic based on the Godot version, but it might be doable.  I didn't re-read what I did to fix the issue.

GUT 9.6 had GUI changes to address Godot 4.6.  I think these changes are backwards compatible, but I did not test it.

# GUT 9.4.1?
If you to craft your own GUT to work in multiple Godot versions I'd definitely consider a point release of a previous version(9.4.1 or 9.5.1).  I'm happy to help if you have any issues.

The [`should_skip_script`](https://gut.readthedocs.io/en/v9.4.0/class_ref/class_guttest.html#class-guttest-method-should-skip-script) virtual method could be useful in your own tests.  This allows you to skip a script (not before it is loaded though), which you could do based on Godot version.  

You could also use a [Pre-Run Hook Script](https://gut.readthedocs.io/en/v9.4.0/Hooks.html) to conditionally add/remove test-scripts/directories based on the Godot version.  I'm not sure how great the documentation is on this, but I know it's possible.

TheDuckCow:
> scripts and editor functionality can be a pain to test. Don't hesitate to provide any other feedback or issues.

A pain indeed, but using gut has been pretty stable in the end. Though tbh _more_ time should be put into test coverage of some features (as one could expect), that's more on me. That being said, test coverage is largely over the functional aspects of code, geometry generation and matching and building blocks like that. A lot of the codebase (and complexity) has to do with user interactions and responding (or rather, fighting) with editor-triggered signals and undo/redo steps vs those the plugin needs to "correct". All that to say, haven't gone too deep into making test coverage of that area, but it's also amongst the easiest to test quickly anyways so it works out. So, no specific wishes from my but I definitely haven't pushed things to the max.

So, back to the issue at hand - and btw, thanks for the detailed response:

> What if the heading was changed to "Valid Godot Versions" and the versions of Godot were listed like this:

Yup that looks perfectly reasonable to me. If there's a known cap, it can be specified. If not, then leave it unbounded until proven otherwise. It's more information that there is today without being dishonest.

> (I have a problem with that...as you can see by the size of this reply and the length of this sub-comment I have put into parenthesis).

I, too, have a problem with this, as you can see from the PR description vs LOC ratio 😄 

> I think this is the only reason why GUT 9.4.0 would not work with Godot 4.5. I don't think there is an actual issue, just that you get a lot of warnings that GUT's Logger is shadowing Godot's Logger. 

Certainly sounds like it's editor warnings. For whatever it's worth, here's at least [one datapoint](https://github.com/TheDuckCow/godot-road-generator/actions/runs/22426170395/job/64934699659) towards "it works" - this run was with gut 9.4.0. At first I was going to say I wouldn't notice errors and warnings anyways since the tests (intentionally) generate a number of errors, due to negative testing. But, I suppose those loggers would also be visible on startup? Perhaps I just missed it.


 > You'd have to do some conditional logic based on the Godot version, but it might be doable.

Hm I wish I could do this more. I also develop addons for Blender and there it's quite easy with the just-in-time-compilation, but more often than not with Godot, the class loads or it doesn't, can't wait for those conditionals downstream (and at the very least, RIP to typing). Nothing is easy!

Appreciate all the brainstorming ideas.

Ultimately, I'm aiming to mess with trying to fix different versions to extend compatibility, sounds like it could be a rabbit hole. If I get some extra time, I'll take another look at hot installing, but barring that, the next best solution would be keeping a git patch that is just checked in applied during the 4.6 runner, containing is the entire diff of gut 9.4.0 to 9.6.0. In fact, that would probably work pretty well, albeit a bit cumbersome.

So, no specific asks at this time - I think the table format proposed is a good change, feel free to take over the branch or close this in favor of doing it fresh your way. And thanks for continuing to support GUT 😎

bitwes:
# The README
I went ahead and just updated the readme directly.  Have a look and let me know if you think it is clear enough now.

# Extra Words
> Ultimately, I'm aiming to mess with trying to fix different versions to extend compatibility, sounds like it could be a rabbit hole.

I don't get many issues for older minor versions of GUT.  It's usually the current version, and dev/beta versions of the next release.  GUT only relies on core aspects of the engine though.  Changes to 2D and 3D systems don't really affect GUT (may affect people's tests, but not GUT).

> here's at least [one datapoint](https://github.com/TheDuckCow/godot-road-generator/actions/runs/22426170395/job/64934699659) towards "it works"

I took a look at the test output.  Per #709, I would have expected to see the Logger warnings at the start of the run.

If you use GUT 9.5 (requiring 4.5.x) you can assert against all the errors you are generating.  It doesn't cleanup the output any, but if a test passes, you can be more sure that the errors in the log are accounted for.  GUT generates some errors, and I always had to read the output of the run just to be sure the errors I saw were the expected ones.  I do that far less now and I'm pretty happy about that.

You will have to add error asserts to all tests that generate an error though, as any unexpected error will cause tests to fail.  You can disable this feature though.


> Not sure how to indicate I am no longer replying to a quote without using a Heading, and I couldn't think of a good heading to add.

Best of luck with your plugin and game.  Great to hear from people using GUT.  If you have any issues in your endeavors, please reach out.

TheDuckCow:
New table looks good! I'll close this PR.

>  It doesn't cleanup the output any, but if a test passes, you can be more sure that the errors in the log are accounted for. 

Yup I'm already not worried about console outputs, I give trust in my assertions in the end since there are code blocks where I want to generate those errors/warnings for the users, and would be cumbersome to try to conditionally push errors based on a live install use case vs gut run.

> I don't get many issues for older minor versions of GUT....

Fair enough, I'll contemplate it.

PR #814, opened 2026-02-27
Avoid error message when opening text files

TopScales:
Whenever a text file, like a README, is opened in the editor, a message is shown saying: "Attempt to call function 'get_base_script' in base 'null instance' on a null instance."

What this PR does, is checking if it is actually a script before checking if it is a test script

### DIFF
--- a/addons/gut/editor_caret_context_notifier.gd
+++ b/addons/gut/editor_caret_context_notifier.gd
@@ -178,6 +178,8 @@ var _scripts_that_have_been_warned_about = []
 var _we_have_warned_enough = false
 var _max_warnings = 5
 func is_test_script(script):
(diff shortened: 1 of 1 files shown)


bitwes:
What version of Godot/GUT are you using?  
How are you opening the files?

I was unable to replicate the error message using Godot 4.6.1 and Gut 9.6.0.  I tried opening .md, .txt, .json and a file with no extension and didn't see the error message.  

I used the File->Open menu and double clicking files in the Filesystem dock.

TopScales:
Was happening in Godot 4.5, but I updated to 4.6 and I don't see the message anymore.

TopScales:
Closing because it seems to be already fixed in last version

PR #822, opened 2026-03-11
Fix failing test when running from cli in debug mode

Curtis-Barnhart:
There are many places in GUT where unused parameters are not prefixed with underscores. I didn't think this was really a big problem, but I ran into one quite unexpected error because of it. When I ran this GUT unit test, with debug mode on from the cli, it actually caused the test to fail, which I do not believe should happen:

Command: `godot -d -s addons/gut/gut_cmdln.gd -- -gconfig= -gtest=./test/unit/test_gut.gd -gunit_test_name=test_can_run_tests_with_parameters`

Error:

```
==============================================
= Run Summary
==============================================

./test/unit/test_gut.gd.TestEverythingElse
- test_can_run_tests_with_parameters
    [Failed]:  Unexpected Errors:
    [1] <engine-1>UNUSED_PARAMETER
    [2] <engine-1>UNUSED_PARAMETER
    [3] <engine-1>UNUSED_PARAMETER
    [4] <engine-1>UNUSED_PARAMETER
    [5] <engine-1>UNUSED_PARAMETER
    [6] <engine-1>UNUSED_PARAMETER
    [7] <engine-1>UNUSED_PARAMETER
          at line 709
```

Running this test from the gui was fine, as well as running it without the `-d` flag, but for some reason running it in debug mode caused an actual failure. I don't understand exactly why this occurs and I'd love to learn more (any ideas?), but regardless it didn't seem like a bad idea to me to change some of the unused parameters to start with underscores. This removes the error and also cleans up a lot of extra warning noise that the unused parameters were generating.

There were also two variables in test_gut.gd that I believe were genuinely not used for anything and could be removed, but it would be good to check that my understanding here is actually correct.

There are possibly many other places with unused parameter warnings that I did not change in this pr - I only changed these particular ones because they were actually causing errors or were in the same file/related files to the ones that were causing errors.

### DIFF
--- a/test/unit/test_doubler.gd
+++ b/test/unit/test_doubler.gd
@@ -128,7 +128,7 @@ class TestDoublingScripts:
 
 	func test_doubling_methods_have_parameters_1():
 		var inst = _doubler.double(DoubleMe).new()
-		assert_source_contains(inst, 'has_one_param(p_one=', 'first parameter for one param method is defined')
+		assert_source_contains(inst, 'has_one_param(p__one=', 'first parameter for one param method is defined')
 
 	# Don't see a way to see which have defaults and which do not, so we default
 	# everything.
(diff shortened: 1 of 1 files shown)


bitwes:
That is odd.  The only thing I can think of is that maybe the editor version is disabling warnings that the command line is not, and some of the warnings are only visible when the debugger is active?  That doesn't really seem right though.

I'm fine with this change but I don't want to merge it until I understand what is going on with the warnings.  I'll keep my eyes open and look into this a bit more when I get a chance.

PR #826, opened 2026-03-12
I803 engine error message

### LINKED ISSUE #803: Error string on engine errors lacks actual error message
# Versions
*(list all versions where you have replicated the bug)*
* Godot: 4.6
* GUT: 9.5
* OS: MacOS

# The Bug

When a test fails due to an engine error, the text that reports why the test failed lacks the original error message from the engine.

Reason is that when there's an engine error, there are 2 strings provided to the OS logger that is registered (in this case GUT's ErrorTracker): `code`, and `rationale`. The `code` is usually some arcane statement like the variable that was asserted on, while the rationale has a more meaningful message. But GUT reports only the code.

# Steps To Reproduce

Run this test:

```gdscript
func test_gut_error_reporting() -> void:
	OS.delay_usec(-10)
	assert_true(true)
```

The test will fail with:

```
    [Failed]:  Unexpected Errors:
    [1] <engine-0>Condition "p_usec < 0" is true.
          at line -1
```

But the actual error message logged by the engine was much better:

```
ERROR: Can't sleep for -10 microseconds. The delay provided must be greater than or equal to 0 microseconds.
```

This is not the best example as even the 'code' message is useful, but often times it isn't. If there's a problem loading a scene, for example, then the error 'code' is just "!valid was true" instead of the much more meaningful string about what part of the scene had a problem.

I believe the fix should be to update the ErrorTracker to do this ternary operation in `get_fail_text_for_errors`, but maybe there's a better format that uses both strings, and maybe there are other locations where this should be done.

```gdscript
	if(errors.items.has(test_id)):
		for error in errors.items[test_id]:
			if(_is_error_failable(error)):
				error_texts.append(str('<', error.get_error_type_name(), '>', error.rationale if error.rationale else error.code)) # <-- use rationale if provided

```

bitwes:
Resolves #803 .

More information is included for engine errors.  The line number, in the test script, where the error occurred is also included in the output.


jack9:
Consider parameterized tests. Some of these changes (and the existing append "\n    ") are going to output unexpected and unreliable results.

PR #834, opened 2026-05-25
Fix syntax for returning dictionary in error_tracker

mcgillij:
found this when running gdformat on my project that was using GUT.


bitwes:
Sure.  Thanks!

PR #851, opened 2026-07-27
Avoid wrapping methods that return Signal

### LINKED ISSUE #833: doubled methods that return a `Signal` cause GUT to hang
# Versions
* Godot: 4.x
* GUT: 9.x
* OS: not OS Specific

# The Bug
When a method that is stubbed to call super returns a `Signal`, GUT will hang forever. 

This happens because all calls to `super` are done using `await` because there is no way to know if the method is a co-routine via the metadata (maybe that has changed, idk).  Since you can't know if the `await` is required, we have to treat everything like it is a co-routine.  

The `await` in the double's wrapper method waits on the returned signal to be emitted, thus causing the run to hang.

This was found when implementing default return values for functions that have a typed return.  GUT should probably generate an error whenever it tries to wrap a method that returns a Signal.  This doesn't help untyped methods that return a Signal, but it's something.  Something should also be added to the documentation.

The workaround is to call `ignore_method_when_doubling` on the class, prior to creating doubles.  In the below example this would look like:

```gdscript
func before_all():
    ignore_method_when_doubling(MyClass, 'get_my_signal')
```

# Steps To Reproduce
The class
```gdscript
class_name MyClass
signal my_signal

func get_my_signal():
    return my_signal
```
The test
```gdscript
extends GutTest

func test_the_problem():
    var inst = partial_double(MyClass).new()
    var result = inst.get_my_signal() # Hangs here
```

JosephDemarest:
## Summary

- detect declared `Signal` return metadata before generating a doubled wrapper
- report an actionable GUT error and leave the method inherited instead of generating an `await` that can stall
- document `ignore_method_when_doubling` and the limitation for untyped Signal-returning methods

## Verification

- focused regression on Godot 4.6.3 / GUT 9.6.1: 2 tests passed, 6 assertions
- complete headless suite: 248 scripts, 1,745 tests, 1,742 passing, 7 existing pending/risky, 3,036 assertions; exit code 0
- GDScript LSP: no errors in the changed source or regression test
- `gdlint` on the new regression test: no problems
- `git diff --check`: clean

Fixes #833

### DIFF
--- a/addons/gut/doubler.gd
+++ b/addons/gut/doubler.gd
@@ -265,8 +265,7 @@ func _create_singleton_double(singleton, is_partial):
 func _stub_method_default_values(parsed):
 	for method in parsed.get_local_methods():
 		if(method.is_eligible_for_doubling() and \
-			!_ignored_methods.has(parsed.resource, method.meta.name) and \
-			!_method_returns_signal(method)):
+			!_ignored_methods.has(parsed.resource, method.meta.name)):
 			_stubber.stub_defaults_from_meta(parsed.script_path, method.meta)
 
 
--- a/addons/gut/doubler.gd
+++ b/addons/gut/doubler.gd
@@ -167,10 +167,27 @@ func _get_singleton_text(parsed, included_methods, is_partial):
 	return src
 
 
+func _method_returns_signal(parsed_method):
+	var return_meta = parsed_method.meta.get('return', {})
+	return typeof(return_meta) == TYPE_DICTIONARY and \
+		return_meta.get('type', TYPE_NIL) == TYPE_SIGNAL
+
+
 func _is_method_eligible_for_doubling(parsed_script, parsed_method):
-	return !parsed_method.is_accessor() and \
-		parsed_method.is_eligible_for_doubling() and \
-		!_ignored_methods.has(parsed_script.resource, parsed_method.meta.name)
+	var method_name = parsed_method.meta.name
+	if(parsed_method.is_accessor() or \
+		!parsed_method.is_eligible_for_doubling() or \
+		_ignored_methods.has(parsed_script.resource, method_name)):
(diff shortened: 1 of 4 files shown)


addons/gut/doubler.gd
```diff
@@ -247,7 +264,9 @@ func _create_singleton_double(singleton, is_partial):
 
 func _stub_method_default_values(parsed):
 	for method in parsed.get_local_methods():
-		if(method.is_eligible_for_doubling() and !_ignored_methods.has(parsed.resource, method.meta.name)):
+		if(method.is_eligible_for_doubling() and \
+			!_ignored_methods.has(parsed.resource, method.meta.name) and \
+			!_method_returns_signal(method)):
```
bitwes: Is this call to `_method_returns_signal` needed?  `is_eligible_for_doubling` already calls it, but I could imagine a case where `is_eligible_for_doubling` returns before checking for the signal return.  I can't think of a scenario where this would matter in this method though.  I removed and all the tests in `test_i833_signal_return.gd` passed.

bitwes (changes requested):
Solid PR.  Thanks.  This implements all the suggested changes mentioned in #833.

See the requested changes.

addons/gut/doubler.gd
JosephDemarest: Removed the redundant `_method_returns_signal` check from `_stub_method_default_values`; Signal rejection remains centralized in `_is_method_eligible_for_doubling`.

The focused `test_i833_signal_return.gd` test passes on Godot 4.6.3 (2/2 tests, 6 assertions). I also reran the full suite: 1,742/1,745 passing with the same seven documented headless pending/risky cases, 3,036 assertions, and exit code 0.

bitwes:
Thanks.

Issue #855, opened 2026-08-06
`assert_almost_eq` with null crashes


CodeSmith32:
# Versions
*(list all versions where you have replicated the bug)*
* Godot: 4.6.1.stable
* GUT: 9.6.1
* OS: Windows 11

# The Bug
When passing `null` to `assert_almost_eq`, GUT crashes, instead of gracefully showing a failed test.

# Steps To Reproduce

```gdscript
extends GutTest

func test_issue() -> void:
    assert_almost_eq(null, 8.0, 0.1)
```

This appears to happen because `GutTest._do_datatypes_match__fail_if_not` allows nulls to pass through.

PR #856, opened 2026-08-06
Add null checks in assert_almost_* methods

CodeSmith32:
- pre-check inputs for nulls before running comparison: report nulls as failed tests vs crashing

This closes issue https://github.com/bitwes/Gut/issues/855
Please feel free to edit this PR / reply here if I'm missing anything.

e.g., I'd assume that nulls in the inputs should just be reported as failed tests vs. crashing, but if not, that's fine. (Though, I'll at least keep these changes in my own project.)

-----

### Note here:

This PR causes passing `null` for `got`, `expected`, or `error_interval` to *either* `assert_almost_eq` *or* `assert_almost_ne` to fail the test. I'd assume both cases are correct, but, to be clear, this does mean that `assert_almost_ne(null, null, 1)` **fails** rather than passes. If that sounds wrong, let me know, or just edit it. Either way, maybe we add a note to the docs for these methods?

Thanks for this amazing tool!

### DIFF
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -977,7 +977,7 @@ func _is_almost_eq(got, expected, error_interval) -> bool:
 
 	return(result)
 
-## assserts got > expected
+## asserts got > expected
 ## [codeblock]
 ##    var bigger = 5
 ##    var smaller = 0
--- a/addons/gut/test.gd
+++ b/addons/gut/test.gd
@@ -201,6 +201,16 @@ func _do_datatypes_match__fail_if_not(got, expected, text):
 	return did_pass
 
 
+# Checks if the input is not null. If it is, a fail occurs. Otherwise, TRUE is
+# returned.
+func _fail_if_null(got, input_name, text):
+	if(got == null):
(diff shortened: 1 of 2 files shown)


bitwes:
Needs some tests.  I got curious about a couple other conditions and came up with the following test that can be expanded later if needed.  Please add the following to `res://test/unit/test_test.gd` under `TestAssertAlmostEq` (around line number 345)

```gdscript
	var datatype_data = [
		[1, 2, null],
		[1, null, 3],
		[null, 2, 3],
	]
	func test_datatype_checks(params = use_parameters(datatype_data)):
		gr.test.assert_almost_eq(params[0], params[1], params[2])
		assert_engine_error_count(0)
````

CodeSmith32:
Thanks @bitwes!

I've added those tests in. I made it a little more thorough:
- I added tests for all permutations of null in the parameters.
- I also wrote a test for different parameter types (not precisely related, but I thought still deserved), and asserted that such tests fail with "Cannot compare..." and don't crash (to prevent regressions)
- Lastly, I added the same tests for `assert_almost_ne`.

I can change/remove these if needed. Just let me know!

